PR #150, opened 2022-05-04
Return extensions and make `describe` visual

aguschin:
This PR uncomments MLEM extension and creates visual representation of `gto describe` resembling Model Card in Studio. WIP example:

image

The model card from the screenshot was printed upon calling `gto describe ml-model`.

### DIFF
--- a/gto/api.py
+++ b/gto/api.py
@@ -1,6 +1,6 @@
 from collections import OrderedDict
 from datetime import datetime
-from typing import List, Optional, Union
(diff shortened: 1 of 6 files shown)


dmpetrov:
What is the motivation to build such UI in CLI? 😄 We don't do that in DVC, Git does not do this. Why MLEM is special?

aguschin:
It was decided we don't need this for now. Closing 👋

Issue #132, opened 2022-04-19
Latest: by SemVer or by timestamp?


aguschin:
`gto which rf prod` should show the latest version of `rf` promoted to `prod`. There are two approaches on selecting the latest one: by SemVer and by timestamp.

E.g. if I promoted `v1.0.0` first, and then promoted `v0.0.1`: latest by SemVer would be `v1.0.0`, while latest by timestamp would be `v0.0.1`.

Promoting non-registered versions makes things more interesting. E.g. if I promoted version from commit `abc1234` and didn't register a version from that commit - the latest by timestamp would be `abc1234` now.

The question is: what should we show by default in Studio and in CLI?

Since we treat SemVer as a number that positively correlates with model quality and stability, the default latest in my mind should be `v1.0.0`. But then if you have a single registered version of model in stage `prod` and then for some reason abandoned registering SemVer versions and just promote, you'll always have that SemVer in `gto latest` and `gto show`. As a more simple example, a user may be confused after promoting `abc1234` because it doesn't appear in `gto show` by default or because `gto latest` doesn't return it.

As an option to resolve this we can keep `--sort-by-semver` default and add flag `--sort-by-time` that will give the recent promoted version/ref.

WDYT?  cc @omesser @dberenbaum @dmpetrov @shcheklein

aguschin:
@omesser, getting back to GTO since MLEM release is on the way 😅 

What you say makes sense to me, and I would like to agree considering the simplicity. But with each "strict" option we need to understand how GTO behaves if a user breaks it. If we don't allow a user to promote a model in a commit without registering it (explicitly in your option1 / implicitly in option2), then what GTO should do if a user creates that git tag by himself? I see these options:

1. Ignore - pretend that promotion didn't happen. Don't show the promotion anywhere. Have a `$ gto doctor` cmd (or `$ gto check-registry-correctness` or anything else) that will let user know that he needs to register a semver.
2. Issue warnings each time user calls some command. (Hey, you have a promotion, but don't have a version, go and register it!)
3. Break each time user calls some command. Ask to register the version to fix this.
4. Automatically register version.

I think 1st and 2nd are preferable. But still, they could be bad since promoting git tag will trigger CI if you push it. So in fact, a user may be confused: I created a tag (though manually), I pushed it, why am I getting some errors?

[...]

PR #162, opened 2022-05-27
Enhance sorting

aguschin:
@amritghimire, please check out this PR. You need `def sort_versions` in `base.py`

Note, that BaseVersion in GTO has `is_registered` and `discovered` atrrs, which I use for sorting also. Not sure you have them already.

`is_registered == True` indicates that it's an explicitly registered SemVer, e.g. `v1.2.3`.
`is_registered == False` indicates that it wasn't explicitly registered, but model in this hexsha was promoted to some stage (e.g. with `model#prod#1` git tag). In this case GTO assumes it's a version, though not explicitly registered. For the version name, GTO uses commit hash.

Considering #132 `is_registered == False` maybe deprecated. But we need a decision there first.

cc @shcheklein: #132 is something I can use your help with.

### DIFF
--- a/gto/base.py
+++ b/gto/base.py
@@ -75,6 +75,42 @@ class BaseArtifact(BaseModel):
         stages = ", ".join(f"'{l}'" for l in self.unique_stages)
         return f"Artifact(versions=[{versions}], stages=[{stages}])"
 
+    @staticmethod
+    def sort_versions(
+        versions,
+        sort=VersionSort.SemVer,
+        ascending=False,
+    ):
+        sort = sort if isinstance(sort, VersionSort) else VersionSort[sort]
+        if sort == VersionSort.SemVer:
+            # sorting SemVer versions in a right way
+            sorted_versions = sorted(
+                (v for v in versions if v.is_registered),
+                key=lambda x: x.version,
+            )[:: 1 if ascending else -1]
+            # sorting hexsha versions alphabetically
+            sorted_versions.extend(
+                sorted(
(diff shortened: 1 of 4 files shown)


gto/base.py
```diff
@@ -58,6 +58,51 @@ def dict_status(self):
...
+            return obj[key]
+        if isinstance(obj, BaseModel):
+            return getattr(obj, key)
+        raise NotImplementedError("Can sort either dict or BaseModel")
+
+    sort = sort if isinstance(sort, VersionSort) else VersionSort[sort]
+    if sort == VersionSort.SemVer:
+        # sorting SemVer versions in a right way
+        sorted_versions = sorted(
+            (v for v in versions if v.is_registered),
```
amritghimire, line 79: @aguschin We don't store is_registered or discovered tag with us. While using the method with dict, `v.is_registered` may throw AttributeError here.

aguschin: Ok, I see [your example repo](https://github.com/iterative/example-mlem-get-started-frozen-v0.1.7/tags) doesn't include these cases, so I think you didn't implement support for `not is_registered` versions.

aguschin: @amritghimire, I've make it infer whether version is SemVer or just a hexsha. Will it work now?

amritghimire: Thanks, it works now.

aguschin: Do you need something else to be implemented now? Otherwise I'll go ahead and create new release.

amritghimire: That is all @aguschin . Thanks for continuous updates to the package.

Issue #156, opened 2022-05-11
Generate huge repo for testing purposes


aguschin:
We need to generate a big repo with 10k+ git tags and 10k+ commits:
- to check how GTO will handle those
- to use in Studio BE for benchmarking and driving the development

If GTO / Studio will process those git tags slow enough, we need to see if we can optimize something:
- maybe we can limit the tags `check-ref` analyzes to only those that reference a single commit
- maybe we can optimize constructing internal registry state in GTO / make it lazier than now (e.g. read stuff by parts/chunks?)

cc @amritghimire

shcheklein:
Can we use https://github.com/iterative/dvcgen ?

aguschin:
Thanks, I'll take a look. @Suor I see you've created the repo, do you think extending it to generate git tags for GTO makes sense?

aguschin:
Ok, I've experimented a bit with `dvcgen` + custom script that generates git tags that register/promote GTO artifacts. No annotations were created, registrations/promotions only.
What I've learnt so far:
1. Rn, GTO is pretty slow.
    - If you have ~100 git tags in your repo, most of the commands will take 0.5-1.5 seconds 🐢 
    - If you have ~1500 git tags in your repo, most of the commands will take 6-12 seconds 🐌 
2. GTO slows down both from gto-created tags and other git tags.
3. Pretty much all commands are slow since they're constructing internal "state of the registry" to validate the action user want to execute (`check-ref` also). E.g. if you want to register new version + bump patch part, GTO needs to read all git tags in repo beforehand 🤔 

Thoughts:
1. This will very soon become tedious for Studio users.
2. Studio BE was correct to use internal representation and storing it in DB.
3. I'll look into this and try to make it faster until the end of the week.

aguschin:
Thanks, @amritghimire! I'll see what I can do. Re your items:
- `get_commit_index` itself neither change anything, nor reconstruct the internal registry state
- got it
- got it

aguschin:
Posting few tools I'll check out to profile GTO
- https://github.com/plasma-umass/scalene
- https://github.com/P403n1x87/austin
- https://github.com/benfred/py-spy

aguschin:
related to https://github.com/iterative/gto/issues/88

PR #163, opened 2022-05-31
Make GTO faster

aguschin:
Addressing findings in #156
Checked this again, it seems and got faster by 50% only. Will see what else can be done, in the future PRs.

### DIFF
--- a/gto/registry.py
+++ b/gto/registry.py
@@ -6,7 +6,12 @@ from git import InvalidGitRepositoryError, NoSuchPathError, Repo
 from pydantic import BaseModel
 
 from gto.base import BasePromotion, BaseRegistryState
-from gto.config import CONFIG_FILE_NAME, RegistryConfig, read_registry_config
+from gto.config import (
+    CONFIG_FILE_NAME,
+    RegistryConfig,
+    assert_name_is_valid,
+    read_registry_config,
+)
 from gto.exceptions import (
     NoRepo,
     VersionAlreadyRegistered,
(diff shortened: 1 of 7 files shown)


gto/registry.py
```diff
@@ -225,13 +232,21 @@ def promote(
...
+        if ref.startswith("refs/heads/"):
+            ref = self.repo.commit(ref).hexsha
         return {
-            "version": self.version_manager.check_ref(ref, self.get_state()),
-            "stage": self.stage_manager.check_ref(ref, self.get_state()),
+            "version": self.version_manager.check_ref(ref, state),
+            "stage": self.stage_manager.check_ref(ref, state),
         }
 
+    def check_ref(self, ref: str):
```
aguschin, line 245: @amritghimire, this is not ideal, but if you create `GitRegistry`, you can use this function body as an example on how to cache state. Looking for better solutions.

amritghimire: This could work. A similar approach for get_stages will also be great. We do checkout to a commit for each get_stages call. But if get_stages use the head when creating GitRegistry, having ref there as well would be better.

aguschin: Why do you call get_stages each time? What stages user sees in MR should depend on config HEAD only. So it doesn't matter what is there in all other commits.

aguschin: E.g. you should always call `$ gto stages` when HEAD of default branch gets updated.

Suor: We need to parse everything for each commit. Because: 
1. We parse incrementally and not all at once
2. Some commits might be removed, in this case we don't want to reparse, we want to recalc.

In perfect world we wouldn't need to create `GitRegistry` and manage it. Things will be cached for us on your side, not sure how hard is this to do though.

Suor: The easiest workaround is probably providing `git_registry` param, i.e. if not `None` then use it instead of recreating each time.

PR #166, opened 2022-06-01
Improve names validation

### LINKED ISSUE #164: Make artifact names, stages, types registry-insensitive
1. Make them case insensitive
2. Restrict them: doesn't allow starting or ending with `-` either, and maybe do not allow starting with a number as suggested by @Suor 
3. It would be helpful if regex or validation func will be exposed in API
4. 2+ symbols

aguschin:
closes #164 

@Suor, do you really need that check exposed in `mlem.api.check_name_is_valid`? It doesn't make sense to the casual user I think, that's why I'm not sure about this. It's accessible in config anyway.

TODO: check that names are correct upon reading/writing to `artifacts.yaml`, as well upon registration/promotions.

### DIFF
--- a/gto/config.py
+++ b/gto/config.py
@@ -22,7 +22,7 @@ CONFIG_FILE_NAME = ".gto"
 
 
 def check_name_is_valid(name):
-    return bool(re.match(r"[a-zA-Z0-9-/]*$", name))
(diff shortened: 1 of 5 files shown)


aguschin:
Ok, merging this for the sake of speed. Feel free to reach me out if you need to change something.

PR #187, opened 2022-06-23
add API to check whether it's a GTO repo

### LINKED ISSUE #186: API call to check if the repo is a GTO repo
There is no clear definition what is "a GTO repo", but I assume:
1. If there is a `.gto` config file - it's a GTO repo
2. If there is an `artifacts.yaml` file - it's a GTO repo
3. If there are some git tags that register/promote artifacts, it's GTO repo

Although all signals may have False Positives (e.g. you created some tags of GTO format with some other tool and didn't mean GTO; or you created `artifacts.yaml` without intention to use it with GTO), I think it's safe enough to expect it's a GTO repo and try to parse it accordingly (at least it's what GTO CLI&API will do).

cc @amritghimire

aguschin:
close #186

### DIFF
--- a/gto/api.py
+++ b/gto/api.py
@@ -19,7 +19,7 @@ from gto.tag import parse_name as parse_tag_name
 from gto.tag import parse_name_reference
 
 
-def is_gto_repo(repo: Union[str, Repo]):
+def _is_gto_repo(repo: Union[str, Repo]):
     """Check if repo is a gto repo"""
     try:
         return GitRegistry.from_repo(repo).is_gto_repo()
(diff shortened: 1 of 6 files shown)


gto/api.py
```diff
@@ -19,6 +19,14 @@
 from gto.tag import parse_name_reference
 
 
+def is_gto_repo(repo: Union[str, Repo]):
+    """Check if repo is a gto repo"""
+    try:
+        return GitRegistry.from_repo(repo).is_gto_repo()
+    except NoRepo:
+        return False
```
aguschin, line 27: Looks a bit strange to me having this in a public API, TBH.

amritghimire (approved):

gto/api.py
aguschin: Making that method private since I think it's some extra case not needed for every GTO API user.

Issue #212, opened 2022-07-15
Make gto compatible with typer~=0.6.0


francesco086:
Continuation of https://github.com/iterative/gto/issues/208

francesco086:
PR that solves the issue: https://github.com/iterative/gto/pull/213

PR #213, opened 2022-07-15
Support typer>=0.6.0

francesco086:
Solves issue https://github.com/iterative/gto/issues/212

I made test_cli.py compatible with typer~=0.5.0, not with typer~=0.4.0 (tests fail in that scenario).

I was wondering if it possible to automatically run the tests for the two different versions on typer. I could not find a good way.

EXTRA: small fix in the README.md (there was an inconsistency with gitignore)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,7 +4,7 @@ from setuptools import find_packages, setup
 
 install_requires = [
     "gitpython",
-    "typer<0.6",
+    "typer",
     "rich",
     "pydantic",
     "ruamel.yaml",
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -4,7 +4,7 @@
 
 install_requires = [
     "gitpython",
-    "typer<0.6",
+    "typer",
```
aguschin, line 7: I checked and it works starting from `0.4.1`
```suggestion
    "typer>=0.4.1",
```

aguschin (approved):
Thank you @francesco086! Merging!

PR #216, opened 2022-07-18
Add deprecation docs

aguschin:


aguschin:
TBC in https://github.com/iterative/gto/pull/219

PR #218, opened 2022-07-20
Deprecate Stages and introduce Labels instead

### LINKED ISSUE #191: Labels versus Stages
We've been discussing this already in https://github.com/iterative/gto/issues/68
Also we recently got a user request of adding Labels mechanics in #179 

## Feedback

I've been asking a couple of question about Stages vs Labels approach in multiple places:

1. In ods.ai Slack I got 5 votes for Labels and 1 vote for Stages.

2. In MLOps.community I got this feedback
image

## Suggestion

Considering answers, I think we need to either
1. Change Stages to Labels (and assume users will implement Stages mechanics by themselves if they want to) (also we have `labels` in `artifacts.yaml` already - how to fix this?)
2. Support both. In this case git tags for Stages could be denoted as `rf:prod` (giving a hint about dicts and 1:1 relation), while Labels could be denoted as `rf#prod`.

WDYT @omesser @dmpetrov ?

aguschin:
This is a continuation of the long discussion re adding Labels. In fact, this still enable Stages workflow, although "Stages" as a term will be deprecated (maybe it will be easier to understand now).
close #191 

Starting with DDD, I'm updating the README to get some feedback first.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,12 +100,21 @@ $ gto unlabel awesome-model prod
 Created git tag 'awesome-model#prod#2!' that removes label 'prod' from 'v0.0.1'
 ```
 
-<details summary="What happens under the hood?">
+<details summary="Some details and options">
 
 GTO creates a special Git tag in a standard format:
 `{artifact_name}#{stage}#{e}!`.
 
 Note, that later you can create this label again, if you need to, by calling `$ gto label`.
+
+You also may want to delete the git tag instead of creating a new one. This is useful if you don't want to keep extra tags in you Git repo, don't need history and don't want to trigger a CI/CD or another downstream system. For that, you can use:
+
+```console
+$ gto unlabel --delete
+Deleted git tag 'awesome-model#prod#1' that added label 'prod' to 'v0.0.1'
+To push the changes upsteam, run:
+git push origin awesome-model#prod#1 --delete
+```
 
 </details>
 
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Git Tag Ops. Turn your Git repository into an Artifact Registry:
 
 - Registry: Track new artifacts and their versions for releases and significant
   changes.
-- Lifecycle Management: Create actionable labels for versions marking status of
+- Lifecycle Management: Create actionable stages for versions marking status of
   artifact or it's readiness to be consumed by a specific environment.
 - GitOps: Signal CI/CD automation or other downstream systems to act upon these
   new versions and lifecycle updates.
@@ -64,15 +64,15 @@ several versions in a given commit, ordered by their automatic version numbers.
 
 </details>
 
-### Create a label
+### Assing a stage
 
-Create an actionable label for a specific artifact version with `gto label`.
-Labels can mark it's readiness for a specific consumer. You can plug in a real
+Assing an actionable stage for a specific artifact version with `gto assing`.
+Stages can mark it's readiness for a specific consumer. You can plug in a real
 downsteam system via CI/CD or web hooks. For example: redeploy an ML model.
 
 ```console
-$ gto label awesome-model prod
-Created git tag 'awesome-model#prod#1' that adds label 'prod' to 'v0.0.1'
+$ gto assign awesome-model prod
+Created git tag 'awesome-model#prod#1' that adds stage 'prod' to 'v0.0.1'
 ```
 
 <details summary="What happens under the hood?">
@@ -82,22 +82,24 @@ GTO creates a special Git tag in a standard format:
 
 The event is now associated to the latest version of the artifact. There can be
 multiple events for a given version, ordered by an automatic incremental event
-number (`{e}`). This will keep the history of your labels creation.
+number (`{e}`). This will keep the history of your stages creation.
 
-Note: if you prefer, you can use simple label tag format without the incremental
-`{e}`, but this will disable the `gto history` command. This is because labeling
-an artifact version where a label tag already existed will require deleting the
-existing tag.
+Note: if you prefer, you can use simple stage tag format without the incremental
+`{e}`, but this will disable the `gto history` command. This is because
+assinging a stage to an artifact version where a stage tag already existed will
+require deleting the existing tag.
 
 </details>
 
-### Remove a label
+### Unassign a stage
 
-Sometimes you need to mark an artifact version no longer ready for a specific consumer, and maybe signal a downstream system about this. You can use `gto unlabel` for that:
+Sometimes you need to mark an artifact version no longer ready for a specific
+consumer, and maybe signal a downstream system about this. You can use
+`gto unassign` for that:
 
 ```console
-$ gto unlabel awesome-model prod
-Created git tag 'awesome-model#prod#2!' that removes label 'prod' from 'v0.0.1'
+$ gto unassign awesome-model prod
+Created git tag 'awesome-model#prod#2!' that unassignes stage 'prod' from 'v0.0.1'
 ```
 
 <details summary="Some details and options">
@@ -105,13 +107,17 @@ Created git tag 'awesome-model#prod#2!' that removes label 'prod' from 'v0.0.1'
 GTO creates a special Git tag in a standard format:
 `{artifact_name}#{stage}#{e}!`.
 
-Note, that later you can create this label again, if you need to, by calling `$ gto label`.
+Note, that later you can create this stage again, if you need to, by calling
+`$ gto assign`.
 
-You also may want to delete the git tag instead of creating a new one. This is useful if you don't want to keep extra tags in you Git repo, don't need history and don't want to trigger a CI/CD or another downstream system. For that, you can use:
+You also may want to delete the git tag instead of creating a new one. This is
+useful if you don't want to keep extra tags in you Git repo, don't need history
+and don't want to trigger a CI/CD or another downstream system. For that, you
+can use:
 
 ```console
-$ gto unlabel --delete
-Deleted git tag 'awesome-model#prod#1' that added label 'prod' to 'v0.0.1'
+$ gto unassign --delete
+Deleted git tag 'awesome-model#prod#1' that assigned 'prod' to 'v0.0.1'
 To push the changes upsteam, run:
 git push origin awesome-model#prod#1 --delete
 ```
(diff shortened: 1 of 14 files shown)


README.md
```diff
@@ -179,6 +205,29 @@ $ gto show churn
...
 
+#### Enabling Stages/Kanban workflow
+
+In some cases, you would like to have a latest label for an artifact version to
+replace all the previous labels. In this case the version will have a single
+label. This resembles Kanban workflow, when you "move" your artifact version
+from one column ("label1") to another ("label2"). This is how MLFlow and some
+other Model Registries works.
+
+To achieve this, you can use `--last-label-for-version` flag (or `--last` for
```
dberenbaum, line 216: Just my take since we discussed this: I don't think the long `--last-label-for-version` flag is needed, and the help description for the flag is probably enough to specify that `--last` means the last label for the version (not the last version for the label).

Also, we use `latest` in other places, so it might be worth considering whether `last` and `latest` have different meanings (in which case we might want to clarify each), or else use one of those words consistently.

README.md
```diff
@@ -265,18 +314,26 @@ $ gto which churn prod --ref
 churn#prod#2
 ```
 
+<details summary="Kanban/Stages workflow">
+If you prefer Kanban/Stages workflow described above, you could use `--last` flag:
+
+```console
+$ gto which churn prod --last
+v3.1.0
```
dberenbaum, line 322: From the example workflow above, I don't see why `gto which churn prod --last` would return `v3.1.0` since `v3.1.0` was never assigned label `prod`.

What happens in this scenario:
1. I assign v3.0.0 to staging.
2. I assign v3.1.0 to staging.
3. I assign v3.1.0 to prod.

What should `gto which churn staging` return? What should `gto which churn staging --last` return? What about `gto show` with and without `--last`?

README.md
```diff
@@ -64,31 +64,57 @@ several versions in a given commit, ordered by their automatic version numbers.
 
 </details>
 
-### Promoting
+### Create a label
```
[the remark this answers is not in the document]
aguschin: I guess i didn't make the intention clear: you can have any number of versions active under a label. The other question is that `gto show` shows only the greatest by default to not overload the screen (the same as it was with earlier - if 2 versions are both in "prod", you'll the greatest semver). There should be a param to control this, like `gto show -n 4` would show you last 4 versions with the label.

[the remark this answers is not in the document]
aguschin: Also, since we have "Labels" in `artifacts.yaml` already, I'm considering keeping "Stages" name. WDYT?

README.md
aguschin: Yes, they have different meanings. I had an idea of introducing two flags: `gto show --greatest` (sort by semver) and `gto show latest` (sort by timestamp) with `greatest` as a default.

Re options: Maybe `--last-stage` or `--latest-stage` could work?

README.md
```diff
@@ -149,15 +175,15 @@ Let's look at the usage of the `gto show` and `gto history`.
...
+and the greatest versions for each label.
 
 ```console
 $ gto show
 ╒═══════════════╤══════════╤════════╤═════════╤════════════╕
 │ name          │ latest   │ #dev   │ #prod   │ #staging   │
 ╞═══════════════╪══════════╪════════╪═════════╪════════════╡
-│ churn         │ v3.1.0   │ -      │ v3.0.0  │ v3.1.0     │
-│ segment       │ v0.4.1   │ v0.4.1 │ -       │ -          │
+│ churn         │ v3.1.0   │ v3.0.0 │ v3.0.0  │ v3.1.0     │
```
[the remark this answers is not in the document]
aguschin: You're right, I didn't update the README completely. Could you please check it now?

README.md
aguschin: > What happens in this scenario

I assume that those are only assignments that were executed to `churn`. then:
```shell
$ gto which churn staging
v3.1.0
$ gto which churn staging --last
v3.0.0
$ gto show
 ╒═══════════════╤══════════╤════════╤═════════╤════════════╕
 │ name          │ latest   │ #dev   │ #prod   │ #staging   │
 ╞═══════════════╪══════════╪════════╪═════════╪════════════╡
 │ churn         │ v3.1.0   │ -      │ v3.1.0  │ v3.1.0     │
 ╘═══════════════╧══════════╧════════╧═════════╧════════════╛
$ gto show --last
 ╒═══════════════╤══════════╤════════╤═════════╤════════════╕
 │ name          │ latest   │ #dev   │ #prod   │ #staging   │
 ╞═══════════════╪══════════╪════════╪═════════╪════════════╡
 │ churn         │ v3.1.0   │ -      │ v3.1.0  │ v3.0.0     │
 ╘═══════════════╧══════════╧════════╧═════════╧════════════╛
```

README.md
dberenbaum: What about `--hide-old-stages`? Anyway, to be clear, it shouldn't block the PR.

README.md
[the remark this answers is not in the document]
dberenbaum: No strong opinion on the name actually, but it seems like there is another possible workflow to consider (again, doesn't need to block this PR). I imagine a typical workflow looks like:

1. Put v1 in dev and deploy to dev env.
2. Move v1 to prod and deploy to prod env (in addition to dev env).
3. Move v2 to dev and deploy to dev env (replacing v1).

When I deploy v2 to the dev env, I no longer want v1 to show up in dev because it has been replaced.

README.md
omesser: @aguschin @dberenbaum 
Suggestion - How about we use a per-label prefix that will make this specific label behave like a stage?
We would be able to support behaviors in parallel:
- free-form labels (user needs to assign and remove)
- stages (assigning a stage overrides previous stages assignments per-artifact-version)
- environments (assigning an environment to an artifact-version will cause this environment to be "unassigned" from other versions of that artifact)

README.md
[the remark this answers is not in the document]
aguschin: Yes, the workflow is valid. Then, I think you either:
1. use `-n 1` flag when GTO outputs multiple versions for a single env (which happens by default now in `gto show`) (which may be OK if you don't need another git tag created for some reason)
2. manually unassign the label with `gto unassign` (which is necessary if you need another git tag created for this unassignment).

README.md
dberenbaum: That summary of the three different behaviors is great @omesser.

TBH I like the current approach by @aguschin where you can switch between those by choosing how to view the labels rather than having it coded into the tag names. It makes it much easier to play around with the different approaches without being locked into one of them. The only part I don't like is that naming/explaining is not that easy, but I don't think it would be easy no matter what approach we take.

omesser: I think the main downside of having this determined by the "querying"/"viewing" command is that you cannot mix and match. This prevents from using GTO to manage environments if you already use it for kanban-like stages and vice versa.
To put it in another way - if I build my workflow and CI to treat gto labels like kanban-stages, it will never make sense to view those as something else. If I use them as deployment environments - it will never make sense to query them as anything else. querying it the wrong way is only room for error, there is not upside to that flexibility after the label mechanism was decided on assignment.

omesser:
@aguschin @dberenbaum 
I keep coming back to this when re-reviewing this PR and discussions;
The free-form-labels (the base behavior in this implementation), while most flexible, seem to be almost useless on its own to the user, and can cause a lot of confusion. I can not think of a use case for this which makes sense, only compromises with env-like or stage-like behavior 🤔 
To recap:
1. env-like - 👍  but with slight overlap with mlem (manageable?)
2. stage-like - 👍  but users not using mlem might still like the above in some scenarios.
3. labels - While technically a base implementation for both the above, on its own not helpful. Also confusing because of the labels in `artifacts.yaml`, which, as a concept is "too close" for comfort.

README.md
aguschin: @omesser, I agree with your argument, but I think flexibility is more important currently for us to see what people want. Also, an important question you raised is how this will behave in Studio when a team collaborates. A misunderstanding may appear when different users interpret the labels with different mechanics (e.g. one with free-form labels and one with stages).

On the second point, I don't think I'll be able to implement this before Studio release, and I doubt it will be handy without Studio BE/FE support cause it just won't be shown there. So let's continue to discuss this and gather some users' feedback to see how they want this to work, and then let's implement this in a separate PR.

aguschin:
Looks like `unassign` takes a bit more time that I expected. I’m going to merge this without `unassign` today, and going to add both `unassign` and `deprecate` a version at once, since they're very similar. `deprecate` an artifact is a bit different, but the approach should be similar, so I'm going to add it also I think.

README.md
```diff
@@ -64,39 +64,71 @@ several versions in a given commit, ordered by their automatic version numbers.
...
+Note: if you prefer, you can use simple stage tag format without the incremental
+`{e}`, but this will disable the `gto history` command. This is because
+assigning a stage to an artifact version where a stage tag already existed will
+require deleting the existing tag.
+
+</details>
+
+### Unassign a stage
+
+Sometimes you need to mark an artifact version no longer ready for a specific
```
aguschin, line 96:
```suggestion
### Unassign a stage

Note: this functionality is in development and will be introduced soon.

Sometimes you need to mark an artifact version no longer ready for a specific
```

README.md
dberenbaum: Seems like we don't have much choice in the short term. In the medium/long term, separate concepts of "stages" and "environments" seem useful.

README.md
```diff
@@ -64,39 +64,73 @@ several versions in a given commit, ordered by their automatic version numbers.
 
 </details>
 
-### Promoting
+### Assing a stage
```
shortcipher3, line 67: Typo, should be `Assign`

README.md
```diff
@@ -64,39 +64,73 @@ several versions in a given commit, ordered by their automatic version numbers.
...
 
-Promote a specific artifact version to a lifecycle stage with `gto promote`.
-Stages can be seen as the status of your artifact, signaling readiness for usage
-by downstream systems, e.g. via CI/CD or web hooks. For example: redeploy an ML
-model.
+Assing an actionable stage for a specific artifact version with `gto assign`.
```
shortcipher3, line 69: Typo, should be `Assign`

shortcipher3 (commented):
I like the flexibility this introduces, it is a better fit for my use case.  Thanks!

README.md
aguschin: Thanks, fixing this in #219

Issue #93, opened 2022-03-24
How to deprecate a model


aguschin:
As one artifact may exist in several branches, it may be not that easy to deprecate it. It does make sense to me - because in different branches there could be a totally separate work.

If branches are temporary, it is no problem. But if you have several permanent branches (e.g. one branch per envs - prod, stage, dev), you either 
1. need to make PRs to all branches
2. need to come up with some convention. E.g. we may consider model "experimental" if it never appeared in the default branch or "probably deprecated" if it appeared, but was removed. 

I tend to assume the (1) would be a way to go - just to avoid creating extra conventions that may be redundant/wrong for some users.

When answering this question we should keep in mind that we should use the same way in Studio and in GTO CLI, otherwise it will be bad UX.

cc @jellebouwman @tapadipti @Suor

aguschin:
@omesser this is what we've been discussing today. I guess we need to understand how this should work and prepare this to be ready for Studio MR phase 2 release. Maybe we need to open another issue since this one is discussing `tag-based`, `branch-based` approaches, while second is not used now.

Issue #145, opened 2022-04-21
`register`: accept (optional) `path` (or new `artifact` cmd)


jorgeorpinel:
> having gto register to also accept path would require (in the current design) to both create git tag AND write to artifacts.yaml, which could be confusing

a) Maybe an optional `path` argument for "advanced" users who want to kill 2 birds with 1 stone

b) Or a new command that combines `register`+`annotate`, kind of like how `dvc pull`=`fetch`+`checkout`. Could be `gto artifact`=`gto register`+`gto annotate` and then we'll see what users prefer based on usage.

_Originally posted by @jorgeorpinel in https://github.com/iterative/gto/pull/136#discussion_r855358442_

PR #219, opened 2022-07-25
Add `unassign` and `deprecate`

### LINKED ISSUE #51: Unclear demotion behavior
Initial idea to introduce `gto demote $MODEL $VERSION $ENV` was based on the assumption that user may want to undo the action of promotion. E.g. if he mistakenly promoted a experimental version to production.

One corner case I found out for this `demote` cmd is this.
1. Let's suppose that you have a `rf` with `v1` and `v2`. Then you promote to production `v2`, then `v1`, then again `v2`.
2. Now you `demote` what's in `production` right now. As a result, `v1` now in `production`.
3. Now you `demote` again. As a result, `v2` now in `production`. But you already tried to `undo` promotion of `v2` to `production`, so it's confusing.

I see two options here:
1. Keep things as they are ^. Maybe there are some user scenarios in which this makes sense.
2. Make another assumption. If you `demote` `v2` from `production`, that means that it shouldn't appear there until you `promote` it again. Then the example above will end up without any version of `rf` promoted to `production`.
3. Disallow `demote` instruction and require to `promote` things instead.
4. Use option 3 + still allow to have `demote` shortcut cmd which will actually execute `promote $PREV_VERSION_IN_THAT_ENV`, but that doesn't look very intuitive in complex scenarios also. As you can't distinguish between `promote` made on purpose and `promote` made because user called `demote`, if you run `demote` 2+ times, you'll stuck between two last versions in that env forever.

I'm leaning towards option 3 right now because it's simpler. Later we can introduce `demote` command as in option 1 or 2, however we decide.

WDYT?
CC @dmpetrov @mike0sv @dberenbaum @omesser @shcheklein

aguschin:
This PR adds
1. remove a stage from the artifact version
2. mark an artifact version as deprecated
3. mark an artifact as deprecated
4. Make Stages work Like Envs

close https://github.com/iterative/gto/issues/93
close #51 

Since we're introducing multiple new actions here, we need to align names and terms in the new context. Maybe
1. `register` and `deregister` an entire artifact 
2. `publish` (`release`?) and `deprecate` a version of the artifact (can switch 1st and 2nd to avoid extra changes)
3. `assign` and `unassign` a stage to a version (or `promote` or `demote` if like the previous names).

WDYT?

A minor note for myself: we need to warn a user that he needs to deprecate a version if that version have active assignments (but we shouldn't do the same when deprecating an artifact: warn about active versions - which is a bit non-symmetrical).

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,8 +47,8 @@ This guide will walk you through the Lifecycle of an artifact. To execute each
 step, you'll need to create a Git tag of a special format, which will trigger CI
 and trigger the job you intend to run:
 
-1. Registering it - marking the very beginning of the artifact history in the
-   Git repository.
+1. Registering artifact - marking the very beginning of the artifact history in
+   the Git repository.
 2. Publishing a version - marking an important change in the artifact.
 3. Assigning a stage to a version - marking the readiness to be consumed by a
    downstream system.
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ Created Git tag `awesome-model@` that registers the artifact.
 
 This can be done to mark the very beginning of the artifact history, so it's
 better to create this tag for the very first commit the artifact was present in
-(even not annotated - see below what an annotation is).
+(even if not annotated - see below what an annotation is).
 
 ### Versioning an artifact
 
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ consumer. You can plug in a real downsteam system via CI/CD or web hooks, e.g.
 to redeploy an ML model.
 
 ```console
-$ gto tag awesome-model --stage prod
+$ gto tag awesome-model --version v0.0.1 --stage prod
 Created git tag 'awesome-model#prod#1'
 ```
 
@@ -95,10 +95,10 @@ require deleting the existing tag.
 
 ### Annotating
 
-So far we've seen how to register new version and assign a stage to an artifact
-versions, but we still don't have much information about them. What about the
-type of artifact (dataset, model, etc.) or the file path to find it in the
-working tree?
+So far we've seen how to register a new version and assign a stage to an
+artifact versions, but we still don't have much information about them. What
+about the type of artifact (dataset, model, etc.) or the file path to find it in
+the working tree?
 
 For simple projects (e.g. single artifact) we can assume the details in a
 downstream system. But for more advanced cases, we should codify them in the
@@ -151,7 +151,7 @@ consumer, and maybe signal a downstream system about this. You can use
 `gto untag` for that:
 
 ```console
-$ gto untag awesome-model --stage prod
+$ gto untag awesome-model --version v0.0.1 --stage prod
 Created git tag 'awesome-model#prod#2!'
 ```
 
@@ -169,7 +169,7 @@ and don't want to trigger a CI/CD or another downstream system. For that, you
 can use:
 
 ```console
-$ gto untag --stage prod --delete
+$ gto untag awesome-model --version v0.0.1 --stage prod --delete
 Deleted git tag 'awesome-model#prod#1'
 To push the changes upstream, run:
 git push origin awesome-model#prod#1 --delete
--- a/README.md
+++ b/README.md
@@ -47,10 +47,10 @@ $ cd example-gto
 
 Registering a version is usually done to mark significant changes to the
 artifact. To release a new version (including the very first one), use
-`gto tag`.
+`gto register`.
 
 ```console
-$ gto tag awesome-model --version v0.0.1
+$ gto register awesome-model HEAD --version v0.0.1
 Created git tag 'awesome-model@v0.0.1' that registers a version
 ```
 
@@ -61,19 +61,19 @@ GTO creates a special Git tag for the artifact version, in a standard format:
 
 The version is now associated to the current Git commit (`HEAD`). You can use
 another Git commit if you provide it's hexsha as an additional argument, like
-`$ gto tag awesome-model abc1234`.
+`$ gto register awesome-model abc1234`.
 
 </details>
 
 ### Assigning a stage to version
 
 To assign an actionable stage for a specific artifact version use the same
-`gto tag` command. Stages can mark the artifact readiness for a specific
+`gto assign` command. Stages can mark the artifact readiness for a specific
 consumer. You can plug in a real downsteam system via CI/CD or web hooks, e.g.
 to redeploy an ML model.
 
 ```console
-$ gto tag awesome-model --version v0.0.1 --stage prod
+$ gto assign awesome-model v0.0.1 prod
 Created git tag 'awesome-model#prod#1' that assigns a stage to 'v0.0.1'
 ```
 
@@ -148,10 +148,10 @@ GTO the artifact file is committed to Git.
 
 Sometimes you need to mark an artifact version no longer ready for a specific
 consumer, and maybe signal a downstream system about this. You can use
-`gto untag` for that:
+`gto deprecate` for that:
 
 ```console
-$ gto untag awesome-model --version v0.0.1 --stage prod
+$ gto deprecate awesome-model v0.0.1 prod
 Created git tag 'awesome-model#prod#2!' that unassigns a stage from 'v0.0.1'
 ```
 
@@ -161,7 +161,7 @@ GTO creates a special Git tag in a standard format:
 `{artifact_name}#{stage}#{e}!`.
 
 Note, that later you can create this stage again, if you need to, by calling
-`$ gto tag` again.
+`$ gto assign` again.
 
 You also may want to delete the git tag instead of creating a new one. This is
 useful if you don't want to keep extra tags in you Git repo, don't need history
@@ -169,7 +169,7 @@ and don't want to trigger a CI/CD or another downstream system. For that, you
 can use:
 
 ```console
-$ gto untag awesome-model --version v0.0.1 --stage prod --delete
+$ gto deprecate awesome-model v0.0.1 prod --delete
 Deleted git tag 'awesome-model#prod#1' that assigned a stage to 'v0.0.1'
 To push the changes upstream, run:
 git push origin awesome-model#prod#1 --delete
@@ -181,10 +181,10 @@ git push origin awesome-model#prod#1 --delete
 
 Sometimes you need mark a specific artifact version as a no longer ready for
 usage. You could just delete a git tag, but if you want to preserve a history of
-the actions, you may find `gto untag` useful.
+the actions, you may again use `gto deprecate`.
 
 ```console
-$ gto untag awesome-model --version v0.0.1
+$ gto deprecate awesome-model v0.0.1
 Created git tag 'awesome-model@v0.0.1!' that deregistered a version.
 ```
 
@@ -194,7 +194,7 @@ If you want to deregister the version by deleting the Git tags itself, you could
 use
 
 ```console
-$ gto untag awesome-model --version v0.0.1 --delete
+$ gto deprecate awesome-model v0.0.1 --delete
 Deleted git tag 'awesome-model@v0.0.1' that registered a version.
 Deleted git tag 'awesome-model#prod#1' that assigned a stage to 'v0.0.1'.
 Deleted git tag 'awesome-model#prod#2!' that unassigned a stage to 'v0.0.1'.
--- a/README.md
+++ b/README.md
@@ -43,7 +43,38 @@ $ git clone https://github.com/iterative/example-gto.git
 $ cd example-gto
 ```
 
-### Versioning
+This guide will walk you through the Lifecycle of an artifact. To execute each
+step, you'll need to create a Git tag of a special format, which will trigger CI
+and trigger the job you intend to run:
+
+1. Registering it - marking the very beginning of the artifact history in the
+   Git repository.
+2. Publishing a version - marking an important change in the artifact.
+3. Assigning a stage to a version - marking the readiness to be consumed by a
+   downstream system.
+4. Unassigning a stage - removing an "artifact is ready for this stage" mark.
+5. Deprecating a version - marking a version as the one that should not be used
+   any longer.
+6. Deregistering an artifact - completing the lifecycle of an artifact and
+   marking it as an outdated one.
+
+### Registering an artifact
+
+Note: this is an optional operation, since publishing the very first version
+with `$ gto publish` already implies registering the artifact.
+
+To start tracking the artifact in the registry, you need to run:
+
+```console
+$ gto register awesome-model
+Created Git tag `awesome-model@` that registers the artifact.
+```
+
+This can be done to mark the very beginning of the artifact history, so it's
+better to create this tag for the very first commit the artifact was present in
+(even not annotated - see below what an annotation is).
+
+### Versioning an artifact
 
 To release a new artifact or a new version, use `gto publish`. This is usually
 done to mark significant changes to the artifact (such as a release).
@@ -230,19 +261,6 @@ git push origin awesome-model@v0.0.1 awesome-model#prod#1 awesome-model#prod#2! 
 
 It looks just the same as `$ gto deprecate awesome-model --delete`, but will
 include all artifact versions that exist.
-
-One more thing: since there is a `$ gto deregister`, there should be the command
-that does the opposite. It's `$ gto register`:
-
-```console
-$ gto register awesome-model
-Created Git tag `awesome-model@` that registers the artifact.
-```
-
-This can be done to undo `deregister` and keep the history or at the very
-beginning to add the artifact to the artifact registry. Although, in the last
-it's not mandatory since publishing the very first version with `$ gto publish`
-already implies registering the artifact.
 
 </details>
 
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ artifact. To release a new version (including the very first one), use
 
 ```console
 $ gto tag awesome-model --version v0.0.1
-Created git tag 'awesome-model@v0.0.1'
+Created git tag 'awesome-model@v0.0.1' that registers a version
 ```
 
 <details summary="What happens under the hood?">
@@ -65,7 +65,7 @@ another Git commit if you provide it's hexsha as an additional argument, like
 
 </details>
 
-### Assigning a stage
+### Assigning a stage to version
 
 To assign an actionable stage for a specific artifact version use the same
 `gto tag` command. Stages can mark the artifact readiness for a specific
@@ -74,7 +74,7 @@ to redeploy an ML model.
 
 ```console
 $ gto tag awesome-model --version v0.0.1 --stage prod
-Created git tag 'awesome-model#prod#1'
+Created git tag 'awesome-model#prod#1' that assigns a stage to 'v0.0.1'
 ```
 
 <details summary="What happens under the hood?">
@@ -152,7 +152,7 @@ consumer, and maybe signal a downstream system about this. You can use
 
 ```console
 $ gto untag awesome-model --version v0.0.1 --stage prod
-Created git tag 'awesome-model#prod#2!'
+Created git tag 'awesome-model#prod#2!' that unassigns a stage from 'v0.0.1'
 ```
 
 <details summary="Some details and options">
@@ -170,7 +170,7 @@ can use:
 
 ```console
 $ gto untag awesome-model --version v0.0.1 --stage prod --delete
-Deleted git tag 'awesome-model#prod#1'
+Deleted git tag 'awesome-model#prod#1' that assigned a stage to 'v0.0.1'
 To push the changes upstream, run:
 git push origin awesome-model#prod#1 --delete
 ```
@@ -196,11 +196,14 @@ use
 ```console
 $ gto untag awesome-model --version v0.0.1 --delete
 Deleted git tag 'awesome-model@v0.0.1' that registered a version.
-Deleted git tag 'awesome-model#prod#1' that assigned a stage.
-Deleted git tag 'awesome-model#prod#2!' that unassigned a stage.
+Deleted git tag 'awesome-model#prod#1' that assigned a stage to 'v0.0.1'.
+Deleted git tag 'awesome-model#prod#2!' that unassigned a stage to 'v0.0.1'.
 To push the changes upstream, run:
 git push origin awesome-model@v0.0.1 awesome-model#prod#1 awesome-model#prod#2! --delete
 ```
+
+This includes all Git tags related to the version: a tag that registered it and
+all tags that assigned stages to it.
 
 </details>
 
@@ -225,12 +228,11 @@ all of them for the artifact. You could do that with
 
 ```console
 $ gto deprecate awesome-model --delete
-Deleted git tag 'awesome-model@registered' that registered an artifact.
 Deleted git tag 'awesome-model@v0.0.1' that registered a version.
-Deleted git tag 'awesome-model#prod#1' that assigned a stage.
-Deleted git tag 'awesome-model#prod#2!' that unassigned a stage.
+Deleted git tag 'awesome-model#prod#1' that assigned a stage to 'v0.0.1'.
+Deleted git tag 'awesome-model#prod#2!' that unassigned a stage to 'v0.0.1'.
 To push the changes upstream, run:
-git push origin awesome-model@registered awesome-model@v0.0.1 awesome-model#prod#1 awesome-model#prod#2! --delete
+git push origin awesome-model@v0.0.1 awesome-model#prod#1 awesome-model#prod#2! --delete
 ```
 
 </details>
--- a/README.md
+++ b/README.md
@@ -128,9 +128,9 @@ git push origin awesome-model#prod#1 --delete
 
 ### Annotating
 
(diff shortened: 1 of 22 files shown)


README.md
```diff
@@ -221,54 +224,85 @@ Add an artifact name to print all of its versions instead:
...
+
+Note: this functionality is experimental and subject to change. If you find it
+useful, please share your feedback in GH issues to help us make it stable.
+
+In some cases, you want to see more than a single version assigned in a stage.
+For that, use `--av` (short for `--assignments-per-version`), e.g. `-1` to show
+all versions.
+
+```console
+$ gto show churn --av -1
```
shcheklein, line 245: @aguschin `--av` - minot style comment. It's not common to have multi-letter shortcuts like this. I would not do this.

aguschin:
Hi, I added `deregister`, `deprecate` and `unassign` to the README. Please check - I think now we need to make sure the terms are meaningful and in the right place, since we're introducing new ones. @shcheklein @omesser I suggest to change some of them, cause I think it makes more sense. Please see and let's discuss.

README.md
```diff
@@ -43,30 +43,61 @@ $ git clone https://github.com/iterative/example-gto.git
...
+Created Git tag `awesome-model@` that registers the artifact.
+```
+
+This can be done to mark the very beginning of the artifact history, so it's
+better to create this tag for the very first commit the artifact was present in
+(even not annotated - see below what an annotation is).
+
+### Versioning an artifact
+
+To release a new artifact or a new version, use `gto publish`. This is usually
```
tapadipti, line 79: Why is this called `publish` instead of `version` or `register-version` or something like that?

README.md
```diff
@@ -43,30 +43,61 @@ $ git clone https://github.com/iterative/example-gto.git
...
+2. Publishing a version - marking an important change in the artifact.
+3. Assigning a stage to a version - marking the readiness to be consumed by a
+   downstream system.
+4. Unassigning a stage - removing an "artifact is ready for this stage" mark.
+5. Deprecating a version - marking a version as the one that should not be used
+   any longer.
+6. Deregistering an artifact - completing the lifecycle of an artifact and
+   marking it as an outdated one.
+
+### Registering an artifact
```
tapadipti, line 61: This is confusing, coz in Studio model registry, you see models that you annotated even if you have not registered them. Should the `Add a model` flow in Studio also call `register` and create the Git tag to register the model? Is it possible to combine these commands somehow, so that the user does not have to use so many commands (register, publish, annotate)?

README.md
```diff
@@ -176,6 +207,63 @@ GTO the artifact file is committed to Git.
 
 </details>
 
+### Deprecating a version
```
tapadipti, line 210: Can we call it `unregister`/`unpublish` a version? So that it is clear that this is the opposite of `register`/`publish` a version? Just like we have `assign` and `unassign`.

README.md
```diff
@@ -176,6 +207,63 @@ GTO the artifact file is committed to Git.
...
+```console
+$ gto deregister awesome-model --delete
+Deleted git tag 'awesome-model@v0.0.1' that published a version.
+Deleted git tag 'awesome-model#prod#1' that assigned a stage.
+Deleted git tag 'awesome-model#prod#2!' that unassigned a stage.
+To push the changes upstream, run:
+git push origin awesome-model@v0.0.1 awesome-model#prod#1 awesome-model#prod#2! --delete
+```
+
+It looks just the same as `$ gto deprecate awesome-model --delete`, but will
```
tapadipti, line 262: How about the first Git tag created while registering (the one without versions)? That will also be deleted, right?

README.md
```diff
@@ -205,54 +293,93 @@ Add an artifact name to print all of its versions instead:
...
-╘════════════╧═══════════╧═══════════╧═════════════════════╧═══════════════════╧══════════════╛
+╒════════════╤═══════════╤══════════════╤═════════════════════╤══════════════╕
+│ artifact   │ version   │ stage        │ created_at          │ ref          │
+╞════════════╪═══════════╪══════════════╪═════════════════════╪══════════════╡
+│ churn      │ v3.1.0    │ staging, dev │ 2022-07-14 10:33:53 │ churn@v3.1.0 │
+│ churn      │ v3.0.0    │ prod         │ 2022-07-09 19:27:13 │ churn@v3.0.0 │
+╘════════════╧═══════════╧══════════════╧═════════════════════╧══════════════╛
+```
+
+#### Enabling multiple versions in the same Stage workflow
```
tapadipti, line 304: We have decided not to support this, right?

README.md
```diff
@@ -205,54 +293,93 @@ Add an artifact name to print all of its versions instead:
...
+╘════════════╧═══════════╧══════════════╧═════════════════════╧══════════════╛
 ```
 
+To enable this workflow, you need to supply the `--av` argument to `gto show`
+and `gto which` commands. Other commands behave the same way regardless of the
+approach you choose.
+
+</details>
+
 #### Enabling Kanban workflow
```
tapadipti, line 331: We decided not to support this, right?

README.md
aguschin: Yes, it's not supported in Studio and it is not how things work by default in GTO. I'm still keep this as options to get users opinion on this earlier. If someone asks about those workflows, we can redirect to this CLI options, and if they find these flags useful, we will have a valuable feedback to design other workflows (either as standalone if we decide to use a special git tag format, or as org-wide options in Studio).

README.md
aguschin: Answered above :)

README.md
aguschin: Yes, it will be deleted if it exists. In this example if doesn't exist. Fixing this! Thanks.

README.md
aguschin: Good idea! Yes, we should do that I think, if we're ok with the names.

README.md
aguschin: `register-version` is a bit long IMO. Could be `version`, what is the pair command? `unversion`?
Then it could be `stage` instead of `assign`. `unstage` sounds ok to me.

README.md
aguschin: Good question. I think it should not. We can explain that artifact is considered to be `registered` if there is any version (that wasn't deprecated/unpublished).

Also I recently understood that what Studio shows is the same you should get if you run GTO on a remote repo, e.g. `$ gto show https://github.com/iterative/model-registry`. GTO doesn't process remote repos yet, but we'll add it soon #220

README.md
```diff
@@ -43,30 +43,62 @@ $ git clone https://github.com/iterative/example-gto.git
...
+1. Registering artifact - marking the very beginning of the artifact history in
+   the Git repository.
+2. Publishing a version - marking an important change in the artifact.
+3. Assigning a stage to a version - marking the readiness to be consumed by a
+   downstream system.
+4. Unassigning a stage - removing an "artifact is ready for this stage" mark.
+5. Deprecating a version - marking a version as the one that should not be used
+   any longer.
+6. Deregistering an artifact - completing the lifecycle of an artifact and
+   marking it as an outdated one.
```
aguschin, line 59: From the minimalistic standpoint, I was thinking to only introduce (and implement) what seams really necessary: 2, 3, 4, 6. But on the other hand, having all commands makes the picture complete (even if some parts are optional). WDYT? Should we for now skip adding 1 and 5 cause they look as not essential, or should we introduce them and structure the README around the most supposedly common commands?

README.md
tapadipti: `version` and `stage` sound good. But since these are words that used more often as nouns than as verbs, they may confuse users. I did suggest `version`, but may be `register-version` or `add-version` is clearer?

README.md
tapadipti: The difference from Studio still is that when you add a model from Studio, we don't call the GTO `register` api and create the model tag without versions. I think we should make this consistent with GTO and also call `register`. Also, IMO having to run 3 commands (register, publish, annotate) for getting your first version up is too much. Can `register` internally call `publish` to register the first version? And can it also accept model description etc and call `annotate` internally?

README.md
tapadipti: I'd suggest having a single command for 1 and annotate. But 5 (unassign version) is good to keep.

README.md
tapadipti: Hmm. Then let's change the text further to indicate that at the moment, we don't really see these as common workflows. Instead of `In some cases, you want to see ...`, we could say `In case / If you want to see ...`.

README.md
aguschin: Then it may be `$ gto publish-version` with an alias `$ gto publish`. Or `register-version` with alias `register`.

README.md
aguschin: Thanks, applying your suggestion!

README.md
```diff
@@ -43,30 +43,61 @@ $ git clone https://github.com/iterative/example-gto.git
...
 ```
 
 <details summary="What happens under the hood?">
 
 GTO creates a special Git tag for the artifact version, in a standard format:
 `{artifact_name}@{version_number}`.
 
-The version is now associated to the current Git commit (`HEAD`). You can have
-several versions in a given commit, ordered by their automatic version numbers.
+The version is now associated to the current Git commit (`HEAD`). You can use
```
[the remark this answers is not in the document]
aguschin: You can do that like this: `$ gto register mymodel --version v0.0.1`. A user could find this if he calls `$ gto register -h`. Do you think it is worth including here?

README.md
aguschin: I agree with the reasoning, but I'm afraid combining `register` and `annotate` in a single command is also confusing. Consider, that `annotate` edits `artifacts.yaml` and don't create a commit with this change, while `register` creates a git tag for an already existing commit. If a single command both changes files in the current index and creates a git tag, that can be quite confusing IMO.

mnrozhkov: Thanks @tapadipti for pointing out this! I would expect that during the [`Add Model`](https://dvc.org/doc/studio/user-guide/model-registry/add-a-model) scenario Studio calls 2 commands 
- `gto register`   -> create a tag 
- `gto annotate` 

Not sure about that a separate command is necessary, but I would upvote for consistent behavior!

aguschin: We at least need a separate `gto register` flow in Studio. May skip separate `gto annotate` flow, if you ask me.

README.md
aguschin: Just had a chat with @mnrozhkov, the suggestion is to remove the `1. Registering an artifact` and use instead `2. Publishing a version`. `1.` doesn't solve any particular problem, and `6. Deregistering an artifact` should be undone by either deleting a deregistering git tag (`mymodel@deregistered`) manually or by `2.`:
```console
$ gto deregister mymodel
Created git tag 'mymodel@deregistered'
# now the model is deregistered with all its versions
$ gto publish mymodel
Created git tag 'mymodel@v3.1.3`
# now the model is registered again
```

mnrozhkov: @aguschin there are few questions for clarification 

- after you publish `mymodel` again, the only 'mymodel@v3.1.3` is active and available in MR for `mymodel`? (old version if any are still deprecated?) 
- can I re-publish a model version that was deprecated? (I don't see any use case for this, just curious) 
- IMO `deprecated` sounds better than `deregister`, also you use verb `register` to explain `publish` command ))) may be it's also not optimal command name?  WDYT?

alex000kim:
my 2c:
- It'll be hard to memorize the differences between `register`, `publish`, `deregister`, `assign`, `unassign`, `depricate`
- Just throwing it out there: would it be possible to use the following commands? (or just a similar pattern)
    - `gto assign awesome-model --version v1.0.0` instead of `gto publish awesome-model --version v1.0.0`
    - `gto unassign awesome-model --version v1.0.0` instead of `gto deprecate awesome-model --version v0.0.1` 
    - `gto assign awesome-model --stage prod` instead of `gto assign awesome-model prod`
    - `gto unassign awesome-model --stage prod` (stays the same) 
    -  `gto archive awesome-model` instead of `gto deregister awesome-model`

README.md
aguschin: Good question about the use case! Please ask any other questions like this, if you have some :)

I guess there are two situations in which you may want to deprecate and then un-deprecate a model:
1. You deprecated an artifact so others don't use it, but then changed your mind (in this case deleting a model deprecation git tag will do, although it don't preserve history)
2. You want to deprecate all previous model versions so others don't use them (e.g. you found a critical bug that seems to exist in all versions prior to some version), but still want to develop and use the artifact further.

in 1 you can just delete a model deprecation git tag, but that won't preserve history of deprecation/undeprecation which could be valuable information (there should have been a reason to deprecate, the user should have been describing that reason in model deprecation git tag message). -- That's why I initially created those commands in pairs (register artifact & deregister artifact).

in 2 you may just go over all previous versions and create a version deprecation git tag for each of them. But that may be a lot of git tags and may not be desirable. Although, this is a premature optimization probably and we should ask user to do this if he wants to deprecate all versions at once, but not an artifact.

On the other topic, I agree that `deprecate` sounds better than `deregister`, but having paired command names may be easier to grasp for the user. Also, we may name this not "register & deregister an artifact", but "add & deprecate an artifact".

tapadipti: > you can just delete a model deprecation git tag, but that won't preserve history

I think we shouldn't implement solutions that don't preserve history

alex000kim: why not though? Git allows you to overwrite history if you really want to.
If a user messes up (makes a typo, pushes a wrong version, etc) they should be able to `delete`/`remove` that tag. 
I am sure users will still find a way to do that with git, so we might as well make it easier for them by supporting deletion behavior with gto.

aguschin:
It's a great idea @alex000kim!
I'm trying to implement this and it seems it should work 🎉  

I'm also checking if the following is possible
```bash
gto assign awesome-model v1.0.0  # instead of `gto publish awesome-model --version v1.0.0`
gto unassign awesome-model v1.0.0  # instead of `gto deprecate awesome-model --version v0.0.1`
gto assign awesome-model prod  # instead of `gto assign awesome-model prod`
gto assign awesome-model v1.0.0 prod  # instead of `gto assign awesome-model prod --version v1.0.0`
gto unassign awesome-model prod
gto archive awesome-model  # instead of `gto deregister awesome-model`
```
if we could remove `--version` and `--stage` options, it could make this even more clear.

README.md
```diff
@@ -43,36 +43,38 @@ $ git clone https://github.com/iterative/example-gto.git
...
+another Git commit if you provide it's hexsha as an additional argument, like
+`$ gto tag awesome-model abc1234`.
 
 </details>
 
-### Assing a stage
+### Assigning a stage
```
alex000kim, line 68: We can elaborate on this e.g. "Assigning a stage to version"

README.md
```diff
@@ -176,6 +144,97 @@ GTO the artifact file is committed to Git.
...
+If you want to deregister the version by deleting the Git tags itself, you could
+use
+
+```console
+$ gto untag awesome-model --version v0.0.1 --delete
+Deleted git tag 'awesome-model@v0.0.1' that registered a version.
+Deleted git tag 'awesome-model#prod#1' that assigned a stage.
+Deleted git tag 'awesome-model#prod#2!' that unassigned a stage.
+To push the changes upstream, run:
+git push origin awesome-model@v0.0.1 awesome-model#prod#1 awesome-model#prod#2! --delete
```
alex000kim, line 202: Could you add a clarification as to why `awesome-model#prod#1` and `awesome-model#prod#2!` are also included?

README.md
```diff
@@ -43,36 +43,38 @@ $ git clone https://github.com/iterative/example-gto.git
...
+To assign an actionable stage for a specific artifact version use the same
+`gto tag` command. Stages can mark the artifact readiness for a specific
+consumer. You can plug in a real downsteam system via CI/CD or web hooks, e.g.
+to redeploy an ML model.
 
 ```console
-$ gto assign awesome-model prod
-Created git tag 'awesome-model#prod#1' that adds stage 'prod' to 'v0.0.1'
+$ gto tag awesome-model --version v0.0.1 --stage prod
+Created git tag 'awesome-model#prod#1'
```
alex000kim, line 77: Could this error message specific the version as well?
E.g. `Created git tag 'awesome-model#prod#1' based on v0.0.1 of awesome-model`

alex000kim: Also, since 1) we can't assign a stage without a version and 2) `gto tag awesome-model --version v0.0.1` is very different behavior than `gto tag awesome-model --version v0.0.1 --stage prod`

What do you think about a `--to-version` option when using `--stage`?

`gto tag awesome-model --stage prod --to-version v0.0.1`

README.md
```diff
@@ -176,6 +144,97 @@ GTO the artifact file is committed to Git.
...
 </details>
 
+### Unassigning a stage
+
+Sometimes you need to mark an artifact version no longer ready for a specific
+consumer, and maybe signal a downstream system about this. You can use
+`gto untag` for that:
+
+```console
+$ gto untag awesome-model --version v0.0.1 --stage prod
```
alex000kim, line 154: same as previous comment: how about this?

`gto untag awesome-model --stage prod --from-version v0.0.1 `

README.md
```diff
@@ -176,6 +144,97 @@ GTO the artifact file is committed to Git.
...
+Note, that later you can create this stage again, if you need to, by calling
+`$ gto tag` again.
+
+You also may want to delete the git tag instead of creating a new one. This is
+useful if you don't want to keep extra tags in you Git repo, don't need history
+and don't want to trigger a CI/CD or another downstream system. For that, you
+can use:
+
+```console
+$ gto untag awesome-model --version v0.0.1 --stage prod --delete
```
alex000kim, line 172: this could be:
`gto untag awesome-model --stage prod --from-version v0.0.1 --delete`

README.md
aguschin: Looks better - easier to read the command. Do you think this should do both version registration and promotion then?
```
$ gto tag x3 --version v0.0.1 --stage prod
Created git tag 'x3@v0.0.1'
Created git tag 'x3#prod#1'
```

alex000kim: I guess whether this makes sense or not depends on whether `gto` should allow promotion to a stage from a git ref OR if only a particular version should be allowed to be promoted to a stage.
I kind of assumed the latter.

aguschin: By default, if you try to assign a stage, but the version doesn't exist for the REF, GTO will register it for you. But if you supply `--skip-registration` git tag, then it won't. 

I.e. let's suppose there are no versions for `x4`:

```
$ gto tag x4 HEAD --stage prod
Created git tag 'x3@v0.0.1'
Created git tag 'x3#prod#1'

$ gto tag x4 HEAD^1 --stage prod --skip-registration
Created git tag 'x3#prod#1'
```

Looks like the difference between `--version` and `--to-version` is this: if you provide `REF` and `--to-version`, you will get an error, cause there should be only one way to specify the Git revision to reference with Git tag you'll create. While if you provide `REF` and `--version`, this will be OK: you register a `version` and reference `REF` with Git tag you create. That's why if you provide `REF`, `--version` and `--stage`, you can both register a version and make an assignment.

README.md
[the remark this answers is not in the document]
aguschin: resolving thread since we decided to implement `tag`/`untag` for now

README.md
aguschin: resolving thread since we decided to implement `tag`/`untag` for now

README.md
alex000kim: Ok, I get the idea, but I am not sure if one command should be allowed to create/modify/delete more than one tag at a time.
I'll defer to others. cc @tapadipti

README.md
tapadipti: @aguschin Sorry for the late reply. Not sure I got you. Do you mean that **in addition to the current Add a model flow**, we need the `gto register` flow in Studio? Or that we don't need `gto annotate` flow? If the latter, then how do we save the model description / labels / etc.?

I agree with @mnrozhkov - I think we need to call both `gto register` and `gto annotate` for consistency. @iterative/studio-backend wdyt?

README.md
tapadipti: - I think it's better to keep the commands simple and not allow doing 2 things (register version and assign stage) with one command.
- And having 2 similar flags (`--version` and `--to-version`) looks confusing to me.
- Also, why do we want to generalize this to `tag`/`untag`, instead of making the commands specifically about versions and stages? Do we see more kinds of tags being required in the future, and want to be able to cover those potential scenarios? Since versions and stages have special meanings for us and since we treat them separately (eg, we only display the latest stage per version but we display all versions per model), I think it's easier to understand and remember the commands if they are separate for the distinct functionalities. Simplifying the name/syntax is a good idea, but not sure if combining them into a single command is.

README.md
aguschin: @tapadipti, you're right - we have `gto register` flow already:
image
Sorry for the confusion :)

README.md
aguschin: > I think it's better to keep the commands simple and not allow doing 2 things (register version and assign stage) with one command.
```
# do you mean we shouldn't allow this:
$ gto tag x5 HEAD --stage prod
Created git tag 'x5@v0.0.1'
Created git tag 'x5#prod#1'

# or this?
$ gto tag x5 HEAD --stage prod --version 1.0.0
```

aguschin: > Also, why do we want to generalize this to tag/untag, instead of making the commands specifically about versions and stages?

From the above comments from @mnrozhkov and @alex000kim I assume it's too hard to memorize 5 different commands and syntaxes for a new user. I agree on this direction: GTO does very simple things and should be a lightweight tool, and having a big tough-to-read README with many commands is what I would like to avoid.

On the other hand, if unifying commands brings more confusion than value, let's not do those. Currently (without `--to-version` and `--from-version`) I think it's pretty straightforward though (maybe it's needed to be clarified in docs): if you supply `--stage`, you do an assignment, otherwise, you do a registration.

README.md
amritghimire: Yes, we can call both annotate and register from Studio when adding model. If we do this, are we to  assume that any model will have at least 1 version registered? 
Should we ignore the models without any version registered from now on?

I am trying to understand the reason to register the model when user may want to just add or annotate a model in artifacts.yaml file at the time.

aguschin: > Should we ignore the models without any version registered from now on?

don't think so.

As I understand this the reason is to simplify user workflow. There are cases when a user would want to annotate and register a version at the same time, so it would be convenient to provide an option to do that. At the same time, someone may want to annotate without registering, that's also fine.

README.md
aguschin: 1. I've implemented `--delete` flags for `untag` and `deprecate` commands (this is not the default behavior). Those delete all relevant Git tags. You can find examples in the README.
2. For `deprecate` I'm implementing the following approach: any new tag created after the artifact was deprecated "undeprecates" it. If you're going to do that, GTO will error out by default, so you'll need to provide `--force` flag.

README.md
aguschin: Iterated on this with @mnrozhkov and eventually decided that
1. `tag` looks cool, but if you try to investigate how it works, it becomes confusing (different behavior with `--stage` and without `--stage`
2. `register` and `assign` are easier cause they separate use cases well
3. `deprecate` can unassign stages and deregister artifacts

I've updated the README and it looks good now: we have 3 commands only: `register`, `assign`, `deprecate`. Almost without breaking changes. Feel free to read and provide any more feedback re this.

README.md
tapadipti: @aguschin Questions:
- In terms of displaying the model in the Studio model registry, how should we distinguish between models that are only annotated and models that are annotated and registered? Should we treat them the same and display both these types of models in the MR?
- In terms of the `Add a model` flow in Studio, should we ask the users if they want to only annotate or annotate&register? I think this will introduce unnecessary confusion to the users. Can we agree on one flow?

aguschin: > how should we distinguish between models that are only annotated and models that are annotated and registered

Please check out [this issue](https://github.com/iterative/gto/issues/226). Maybe this is a solution.

on a side note, a similar thing in GTO was proposed by @jorgeorpinel a while ago https://github.com/iterative/gto/issues/145

README.md
aguschin: I've implemented this. So now:
- in CLI you have `register`, `assign`, `deprecate`.
- in API you have `register`, `unregister`, `assign`, `unassign`, `deprecate`

This is because Studio is the main user of API, and having separate commands for them is clearer.
For human in CLI we'll have `deprecate` unifying `deregister` and `unassign`, making things simpler, hopefully.

A side note: it looks like API and CLI in GTO diverge more and more over time.

aguschin:
I'm merging this with how it works to release existing changes. Let's enhance on top of what it's there already. Follow-up issues
- https://github.com/iterative/gto/issues/226
- https://github.com/iterative/gto/issues/227
- https://github.com/iterative/gto/issues/228

Feel free to join any discussion if you wish.

README.md
aguschin: @tapadipti, do we have an issue for this already or should we create one?

tapadipti: No we don't. I'm reading https://github.com/iterative/gto/issues/226 and it looks like we should conclude that discussion and then create required tickets in Studio. Has there been any discussion on that?

aguschin: No, nothing as far as I know.

PR #225, opened 2022-08-17
`annotate`: Introduce `requirements`.

daavoo:
Quick and dirty draft.
Being discussed in https://github.com/iterative/studio/issues/4118

### DIFF
--- a/gto/api.py
+++ b/gto/api.py
@@ -55,6 +55,7 @@ def annotate(
     must_exist: bool = False,
     labels: List[str] = None,
     description: str = "",
+    requirements: List[str] = None,
     # update: bool = False,
 ):
     """Add an artifact to the Index"""
@@ -65,6 +66,7 @@ def annotate(
         must_exist=must_exist,
         labels=labels,
         description=description,
+        requirements=requirements,
         update=True,
     )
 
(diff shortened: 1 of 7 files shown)


gto/config.py
```diff
@@ -26,6 +26,10 @@ def check_name_is_valid(name):
     return bool(re.match(r"[a-z][a-z0-9-/]*[a-z0-9]$", name))
 
 
+def check_requirement_is_valid(name):
+    return bool(re.match(r".+:\s\d", name))
```
daavoo, line 30: What should be the expected format for `--requirements`? 
Should we enforce a pattern or not?

I am forcing `{name}: {version}` here

gto/index.py
```diff
@@ -126,7 +130,15 @@ def write_state(self, path_or_file: Union[str, IO]):
...
-        self, name, type, path, must_exist, labels, description, update
+        self,
+        name,
+        type,
+        path,
+        must_exist,
+        labels,
+        description,
+        update,
+        requirements=None,
```
daavoo, line 141: Tried to add it in backward-compatible way to don't break existing calls, not sure if this methods are "public"

gto/api.py
```diff
@@ -55,6 +55,7 @@ def annotate(
     must_exist: bool = False,
     labels: List[str] = None,
     description: str = "",
+    requirements: List[str] = None,
```
daavoo, line 58: typer support for dict is non-existing (as far as I can tell from quick look). Should we parse list from CLI but convert to key-value internally?

PR #239, opened 2022-08-31
Make artifacts without Git tags "annotated, but not registered"

### LINKED ISSUE #226: Consider `registered` only those artifacts that have git tags
Currently, we consider `registered`:
1. All artifacts that have git tags
2. All artifacts that have annotations in `artifacts.yaml` (not in all commits. For GTO: in HEAD only. For Studio: in branches' HEADs only).

I noticed multiple times that this definition is a source of confusion (at least in conversations with @tapadipti @mnrozhkov @omesser). It's hard to explain and moreover, what's registered depends on the check outed commit in GTO (means it's different for different commits).

Suggestion: remove the 2nd clause and consider artifacts that have annotation in `artifacts.yaml`, but don't have git tags, `unregistered`.

(Besides, `register` an artifact is often confused with `register` a version. We may need to resolve this somehow.)

If we do this,
- in Studio we'll have `registered` artifacts and `unregistered` which are "candidates" to become registered (may use some other naming).
- In GTO, `$ gto show` won't show `unregistered` artifacts, but `$ gto show --unregistered` (or something) will.

aguschin:
Close #226 
Now the artifact without Git tags, but with annotation, won't be considered as "registered artifact".
Those artifacts will be marked with `*` in `gto show` (and show be also marked differently in Studio):
```
$ gto show
╒════════════╤══════════╤════════╤═════════╤════════════╕
│ name       │ latest   │ #dev   │ #prod   │ #staging   │
╞════════════╪══════════╪════════╪═════════╪════════════╡
│ churn      │ v3.1.0   │ v3.1.0 │ v3.0.0  │ v3.1.0     │
│ segment    │ v0.4.1   │ v0.4.1 │ -       │ -          │
│ cv-class   │ v0.1.13  │ -      │ -       │ -          │
│ *annotated │ -        │ -      │ -       │ -          │
╘════════════╧══════════╧════════╧═════════╧════════════╛
```

We still need to come up with some term that won't be confusing: `register an artifact` and `register a version` are usually confused with each other. Maybe instead of `unregistered artifacts` (those that are annotated, but don't have Git tags yet) they should be called `artifact candidates`, or `register a version` should become `publish a version`. WDYT @tapadipti @mnrozhkov?

### TODO
- [ ] update README about this `*`

### DIFF
--- a/README.md
+++ b/README.md
@@ -258,8 +258,9 @@ $ gto show
 ╘═══════════════╧══════════╧════════╧═════════╧════════════╛
 ```
 
-Here we'll see both artifacts that have Git tags only and those that are
-annotated only in `artifacts.yaml` (the latter will be mark with asteriks, e.g.
+Here we'll see artifacts that have Git tags or are annotated in
+`artifacts.yaml`. The artifacts that have annotation, but have no Git tags, are
+considered yet `unregistered` and will be marked with an asterisk, e.g.
 `*annotated`. Use `--all-branches` or `--all-commits` to read `artifacts.yaml`
 from more commits than just `HEAD`.
 
--- a/README.md
+++ b/README.md
@@ -258,9 +258,10 @@ $ gto show
 ╘═══════════════╧══════════╧════════╧═════════╧════════════╛
 ```
 
-Here we'll see both artifacts that have Git tags only and those annotated in
-`artifacts.yaml`. Use `--all-branches` or `--all-commits` to read
-`artifacts.yaml` from more commits than just `HEAD`.
+Here we'll see both artifacts that have Git tags only and those that are
+annotated only in `artifacts.yaml` (the latter will be mark with asteriks, e.g.
+`*annotated`. Use `--all-branches` or `--all-commits` to read `artifacts.yaml`
+from more commits than just `HEAD`.
 
 Add an artifact name to print all of its versions instead:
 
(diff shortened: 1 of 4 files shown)


tapadipti:
> We still need to come up with some term that won't be confusing: register an artifact and register a version are usually confused with each other.

Registering an artifact can be the same as registering the first version of the artifact, right? (Unless the user does not specify a version number). So, registering a version is a matter of specifying the version number only, right?

> Maybe instead of unregistered artifacts (those that are annotated, but don't have Git tags yet) they should be called artifact candidates, or register a version should become publish a version.

I think "unregistered" is clear. "Artifact candidates" could be not just the annotated-but-unregistered artifacts but, in the future, DVC pipeline outputs also, right? IMO, we can call these annotated-but-unregistered artifacts simply "unregistered" and indicate that they are candidates for registration, and that they can register them by using the `gto register` command with or without a version number.

aguschin:
> Registering an artifact can be the same as registering the first version of the artifact, right? (Unless the user does not specify a version number). So, registering a version is a matter of specifying the version number only, right?

Not quite. Let's suppose you just created an Git tag named `rf#prod` (in a repo without Git tags). There is no explicit SemVer version registered for `rf`, although it can be understood as a `registered` artifact (if we count any git tags for an artifact). I will check this with @dmpetrov, but this sounds to me like a situation where we need to assume the artifact is registered. 

> could be not just the annotated-but-unregistered artifacts but, in the future, DVC pipeline outputs also, right?

Yes! Good point.

PR #267, opened 2022-09-16
Mention the use of the '--force' flag when raising WrongArgs error

### LINKED ISSUE #266: Hint using '--force' when assignment already exists
Currently when using `gto assign` to assign a stage to an artifact's version that has already happened, GTO return an error like:

```
❌ Version 'v3.0.0' is already in stage 'prod'
```

It might be helpful to include option to use `--force` in this message, as it might be what a user wants to do, turning the message into something like:

```
❌ Version 'v3.0.0' is already in stage 'prod'. Use the '--force' flag to reassign the 'prod' stage to version 'v3.0.0'
```

jellebouwman:
Closes #266

### DIFF
--- a/gto/registry.py
+++ b/gto/registry.py
@@ -314,7 +314,7 @@ class GitRegistry(BaseModel):
             and any(vstage.stage == stage for vstage in found_version.get_vstages())
         ):
             raise WrongArgs(
-                f"Version '{found_version.version}' is already in stage '{stage}'"
+                f"Version '{found_version.version}' is already in stage '{stage}. Use the '--force' flag to reassign the '{stage}' stage to version '{found_version.version}"
             )
         # TODO: getting tag name as a result and using it
         # is leaking implementation details in base module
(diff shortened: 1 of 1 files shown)


aguschin (approved):

aguschin:
Thank you @jellebouwman! Good idea. Merging!

gto/registry.py
```diff
@@ -314,7 +314,7 @@ def assign(  # pylint: disable=too-many-locals
             and any(vstage.stage == stage for vstage in found_version.get_vstages())
         ):
             raise WrongArgs(
-                f"Version '{found_version.version}' is already in stage '{stage}. Use the '--force' flag to reassign the '{stage}' stage to version '{found_version.version}"
+                f"Version '{found_version.version}' is already in stage '{stage}. Use the '--force' to create new Git tag nevertheless."
```
jellebouwman, line 317: I'm all for Star Wars references - but should it be:

`Use the '--force' flag to create a new Git tag nevertheless.`

aguschin: ahh, merged it too early

Issue #25, opened 2022-02-21
Make `show` and other commands work with remote repos


aguschin:
Right now they work on local repos only, that's a limitation of `gitpython`.

francesco086:
Hi, I'd like to contribute to this one. Any suggestion on where to start from?

aguschin:
Thanks for asking!

There are few parts to this:
1. read-only operations (`show`, `describe`, `check-ref`, `history`, `stages`). Specifically:
  1a. read Git tags (`check-ref`; `history` and `show` without commits)
  1b. read commits (`describe`)
  1c. read both (`show --all-commits`, `history`, `stages`)
2. write operations (`assign`, `register`, `deprecate`, `annotate`, `remove`).  Specifically:
  2a. create/remove git tag(s) and push (`assign`, `register`, `deprecate`)
  2b. create commit and push (`annotate`, `remove`) 

AFAIK, the only way to support this in full is [to clone the repo locally to temp dir first](https://stackoverflow.com/questions/1178389/browse-and-display-files-in-a-git-repo-without-cloning). In this case, the ideal way would be to contribute to scmrepo in the issue above, I suppose, considering DVC already have something like this implemented.

There are also an option to support 1a partially. The thing is, you can read Git tags from a remote repo with `git` CLI like this: `git ls-remote --tags https://github.com/iterative/example-gto`. Didn't check if this is supported in GitPython. The justification for this approach [is this PR](https://github.com/iterative/gto/pull/239) - i.e. all artifacts that don't have Git tags are considered unregistered and are optional to show in `gto show`.

This partial option is good for `gto show` command without `--all-commits` and `--all-branches` flags, and won't enable `gto describe`, `gto history`, etc. Because it should be much quicker than fully cloning repo, eventually we'll need to support both I suppose.

aguschin:
Found a similar explanation I've posted in another issue earlier https://github.com/iterative/gto/issues/220#issuecomment-1198236090

francesco086:
Thanks @aguschin !
I would surely start from the read-only operation.

Reading through what you sent me, I found this especially illuminating: https://stackoverflow.com/a/1179728/19782654
git is distributed, however often times (always?) there is a reference remote repo. One could choose to rely on those and use GitLab/GitHub APIs.
Otherwise there is no way around, one need to do a git clone.
For tags only, one could also rely on `git ls-remote` as you suggested.

[...]

PR #269, opened 2022-09-16
gto show --repo accepts remote repositories

francesco086:


francesco086:
There are issues with windows not managing temporary files / directories properly (see e.g. [this](https://github.com/iterative/gto/actions/runs/3068405644/jobs/4955833273#step:8:132).
However, with more recent python versions, the issue seems to have been solved.

I found [this](https://bugs.python.org/issue42796) discussion.

I tried to find a workaround but was not able to, also because I have no windows and for each attempt I need to push and wait for CI results.

What's our next best action? Any developer with windows who can help? Or shall we just exclude windows people with python < 3.9 ?

aguschin:
Good stuff @francesco086 ! Confirm it works on my Mac!

Re windows tests, let's not implement it. If someone on windows will try running it and it will fail, he will bring this to issues and we will consider implementing this. About tests, let's skip the test execution. Something like this should work:

```
# conftest.py

def skip_matrix(skip: Sequence[Tuple[str, str]]):
    current_os = os.environ.get("GITHUB_MATRIX_OS")
    current_python = os.environ.get("GITHUB_MATRIX_PYTHON")
    if (current_os, current_python) in skip:
        return pytest.mark.skip(
            reason=f"This test is only for {os_} and python:{python}"
        )
    return lambda f: f

# test_file_with_failing_test.py

skip_remote_tempdir_matrix = skip_matrix(
    [("windows-latest", "3.7"), ("windows-latest", "3.8")]
)

@skip_remote_tempdir_matrix
def failing_test():
    ...
```
If you need a working example, please [see MLEM](https://github.com/iterative/mlem/blob/main/tests/contrib/test_heroku.py#L36) (but note, that in MLEM we wanted to run a test in a single CI job - py3.7 on ubuntu, and thus it's the "skip wrapper" is different).

Could we also make this work? It works if you add `.git` in the end, but would be cool if it worked even without it :)
```shell
gto show -r https://github.com/iterative/example-gto
```

aguschin:
Cool! I see you almost fixed everything. I will review the code tomorrow (right now it's the late evening for me). Meanwhile, do you want to add support for other read-only CLI commands in this PR? Like for `check-ref`, `history` and `stages`?

francesco086:
Hi @aguschin ! 

I went crazy with the windows thingy... I wanted to add a simple message in the error: "Are you using windows with python < 3.9? This may be the reason of this error: https://bugs.python.org/issue42796. Consider upgrading python."
To do this, long story short, I was driven crazy. In windows with python 3.8, whereas an error was indeed raised, the pytest assert raise did not recognize it. No idea why! The fact that each time I had to push to github and wait for CI made this trivial thing really lengthy. Eventually I resorted to give up, and test that this extra message is sent only with windows and python 3.7.

Regarding the skip, I preferred the `pytest.mark.skipif` over your custom `skip_matrix`.

Besides that, I added the support for url without the `.git` at the end (had only to change the regex).

Waiting for your code review. I took a lot of liberty in how I wrote and structured the code. Probably too much, it looks too inconsistent with the pre-existing code. Well... for tomorrow!

I would suggest to first merge this one, and then extend to other commands? If you don't mind, I find it easier this way. Let's first get this chunk done.

gto/git_utils.py
```diff
@@ -0,0 +1,71 @@
+import inspect
+import logging
+import re
+from functools import wraps
+from tempfile import TemporaryDirectory
+from typing import Callable, Dict
+
+from git import Repo
+
+# TODO: make a new type out of Union[str, Repo]
```
aguschin, line 10: Do you want to address this in this or next PR?

francesco086: oh! I thought I removed it eventually :)

At first I thought it was a good idea, then reconsidered it. Surely it's a bit redundant to see `Union[str, Repo]` in so many places, but on the other hand for a user it's more understandable than a new type, because he will have to look up for its definition.

So, I'd suggest to remove the TODO, if you agree.

gto/git_utils.py
```diff
@@ -0,0 +1,71 @@
...
+    ) -> Dict[str, object]:
+        kwargs_complement = {
+            k: args[i]
+            for i, k in enumerate(inspect.getfullargspec(f).args)
+            if k not in kwargs.keys() and i < len(args)
+        }
+        kwargs.update(kwargs_complement)
+        return kwargs
+
+    return wrapped_f
```
aguschin, line 52: This should work since repo is always the first arg to all CLI functions, and is more concise:

```suggestion
def git_clone_if_repo_is_remote(f: Callable):
    @wraps(f)
    def wrapped_f(repo, *args, **kwargs):
        if isinstance(repo, str) and is_url_to_remote_repo(repo=repo):
            try:
                with TemporaryDirectory() as tmp_dir:
                    logging.debug("create temporary directory %s", tmp_dir)
                    git_clone(repo=repo, dir=tmp_dir)
                    return f(tmp_dir, *args, **kwargs)
            except (NotADirectoryError, PermissionError) as e:
                raise e.__class__(
                    "Are you using windows with python < 3.9? "
                    "This may be the reason of this error: https://bugs.python.org/issue42796. "
                    "Consider upgrading python."
                ) from e
            logging.debug("temporary directory %s has been deleted", tmp_dir)
        else:
            return f(tmp_dir, args, **kwargs)

    return wrapped_f
```

francesco086: Mmm... I wrote the tests to make sure that the decorator works even if repo is not the first argument. I find it very much prone to problems if you create a decorator with such assumption.

So, currently your code suggestion breaks the tests even after fixing line 30 to `return f(repo, *args, **kwargs)`.

Are you sure you want to make the decorator less general? How would you suggest to make clear that the repo argument must be the first argument in the decorated function?

tests/git_utils/test_git_clone_if_repo_is_remote.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+@only_for_windows_py_lt_3_8
+def test_if_repo_is_remote_url_and_windows_os_error_then_hint_win_with_py_lt_3_9_may_be_the_cause():
+    with pytest.raises(OSError) as e:
+        decorated_func(repo=get_example_http_remote_repo(), spam=0, jam=3)
+    assert e.type in (NotADirectoryError, PermissionError)
+    assert "windows" in str(e)
+    assert "python" in str(e)
+    assert "< 3.9" in str(e)
```
aguschin, line 45: Why not check for the exact error message in one line? Just curious if there is any reason.

tests/git_utils/test_git_clone_if_repo_is_remote.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+@git_clone_if_repo_is_remote
+def decorated_func(
+    spam: int, repo: Union[Repo, str], jam: int
+):  # pylint: disable=unused-argument
+    return repo
```
aguschin, line 52: I think this one will fail if you accept the suggestion I did above (cause `repo` is not the first arg here).

tests/git_utils/test_git_clone_if_repo_is_remote.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def assert_f_called_with_repo_return_repo(repo: Union[str, Repo]) -> None:
+    assert decorated_func(0, repo, 3) is repo
+    assert decorated_func(0, repo, jam=3) is repo
+    assert decorated_func(0, jam=3, repo=repo) is repo
+    assert decorated_func(spam=0, jam=3, repo=repo) is repo
```
aguschin, line 59: Again, if you accept that change, this is no longer needed, I assume.

tests/git_utils/test_is_url_to_remote_repo.py
```diff
@@ -0,0 +1,22 @@
...
+        get_example_http_remote_repo(),
+        get_example_http_remote_repo_without_dot_git_suffix(),
+    ),
+)
+def test_if_remote_url_then_true(repo: str):
+    assert is_url_to_remote_repo(repo=repo) is True
+
+
+def test_if_local_url_then_true():
+    assert is_url_to_remote_repo(repo="/local/path") is False
```
aguschin, line 22: Maybe add a couple more examples with parametrize? Like `.` and relative path, e.g. `local/path`

tests/skip_presets.py
```diff
@@ -0,0 +1,16 @@
+import pytest
+
+from tests.utils import (
+    is_os_windows_and_py_lt_3_8,
+    is_os_windows_and_py_lt_3_9,
+)
+
+skip_for_windows_py_lt_3_9 = pytest.mark.skipif(
```
aguschin, line 8: The same q as above, does it make sense to keep these in `conftest.py`? Except for introducing one more file. @mike0sv ?

tests/utils.py
```diff
@@ -32,3 +33,17 @@ def check_obj(
         obj_values = omit(obj, skip_keys)
     values = omit(values, skip_keys)
     assert_equals(obj_values, values)
+
+
+def is_os_windows_and_py_lt_3_9() -> bool:
+    return (
+        os.environ.get("GITHUB_MATRIX_OS") == "windows-latest"
+        and os.environ.get("GITHUB_MATRIX_PYTHON", "2") < "3.9"
```
aguschin, line 41: Why `"2"`?

aguschin (changes requested):
Looks good overall! Please see some comments and suggestions I had @francesco086

gto/git_utils.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+def git_clone(repo: str, dir: str) -> None:
+    logging.debug("clone %s in directory %s", repo, dir)
+    Repo.clone_from(url=repo, to_path=dir)
```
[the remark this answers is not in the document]
francesco086: I did it in this way at first :) but then realized: we are now writing some interface function that should eventually work even if change git tool. Currently we are using `gitpython` but let's say that we then will need to switch. Perhaps there won't be any other choice but to call `git` from the shell. Then it is better to keep the `git_clone` function interface not returning anything (could return the path to the git dir, but that's an input given, it does not make sense to return it).

One additional argument. As far as I know, it is good practice to separate queries from commands. Commands change the state of the machine, and should not return anything. And `git_clone` is a command...

gto/git_utils.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+def is_url_to_remote_repo(repo: str) -> bool:
+    # taken from https://stackoverflow.com/a/22312124/19782654, modified to include url without .git at the end
+    REGEX_REMOTE_GIT_REPO = (
```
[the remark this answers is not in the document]
francesco086: good idea!

tests/api/data/__init__.py
```diff
@@ -0,0 +1,15 @@
+import json
```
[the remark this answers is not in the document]
francesco086: I wanted to ask you about this :)

I noticed that you don't separate clearly the tests by function under test. I find it essential to have maintainable tests.
Another option would be to organize tests in classes with the name of the function under test.
I prefer the way I have done because it allows me also to have a scoped (e.g. `api`) data module, and I find it easier to navigate through files than scrolling long files.

But you are the code maintainer, so the choice is yours, I just wanted to show you an alternative :)

If you do like it, I can re-structure all tests in `tests/test_api.py` accordingly. Otherwise I will make the tests I wrote fit in the pre-existing structure.

tests/git_utils/data/__init__.py
```diff
@@ -0,0 +1,6 @@
+def get_example_http_remote_repo() -> str:
+    return "https://github.com/iterative/example-gto.git"
+
+
+def get_example_http_remote_repo_without_dot_git_suffix() -> str:
+    return "https://github.com/iterative/example-gto"
```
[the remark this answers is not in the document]
francesco086: Potentially a constant in a module can be overwritten using `global` and get to a mess.

Of course one could also use the same procedure to overwrite the `get` function, but it should feel very very wrong for any developer :)

I just got used to this style, but I am totally fine switching them to constants, if you prefer :) shall I do it?

tests/git_utils/test_git_clone_if_repo_is_remote.py
francesco086: Readability (imho) - Sparse is better than dense

tests/git_utils/test_git_clone_if_repo_is_remote.py
francesco086: indeed, need to come to a decision on your comment above and in case I will adjust accordingly

tests/git_utils/test_is_url_to_remote_repo.py
francesco086: good idea!

tests/utils.py
francesco086: So: can't leave it empty or use None, because python raises an issue (can't compare None to a string - I was surprised that python was so smart to check that out!).

I set "2" to make the condition true in case the python version cannot be detected (if I am not sure, better skip it).

I was not totally happy about it, I would be glad to hear a suggestion.

gto/git_utils.py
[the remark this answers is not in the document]
aguschin: Ok, got your argument. Since it's internals not exposed to users, I think we could do the either way (return something that can change or not return it), so it's up to you.

The other question - maybe move `with TemporaryDirectory() as tmp_dir` here and make this function itself a context manager? Like this https://github.com/iterative/gto/blob/0293e674593293f76ee5ad288dcb5101edc8415f/tests/git_utils/test_git_clone_if_repo_is_remote.py#L63

Then you'll split the cloning logic from the wrapper logic.

tests/api/data/__init__.py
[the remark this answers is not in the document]
aguschin: TBH, I would refrain from refactoring all the tests into anything :) If you mean tests where I test couple of related commands (add+remove, register+deprecate), I do that because those are connected. I could write fixtures OFC, but then the number of repetitions would be greater I think (like when you need to write a test + a fixture that does roughly the same). You can think of it like preferring integration tests to unit tests for those cases, if that's analogy is correct here :)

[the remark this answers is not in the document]
aguschin: If you wish to do it, we can discuss it OFC. But in this case I need more details to better understand WDYM. Do you mean creating a separate files for each command? Like `test_show.py`, `test_annotate.py`, etc?

tests/git_utils/data/__init__.py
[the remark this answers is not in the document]
aguschin: Yes please :)

gto/git_utils.py
francesco086: done

tests/utils.py
aguschin: Hm, I see. Nothing comes to mind except for this I guess
```suggestion
        and os.environ.get("GITHUB_MATRIX_PYTHON", "") < "3.9"
```

gto/git_utils.py
```diff
@@ -0,0 +1,65 @@
...
+
+        return result
+
+    def _turn_args_into_kwargs(
+        args: tuple, kwargs: Dict[str, object]
+    ) -> Dict[str, object]:
+        kwargs_complement = {
+            k: args[i]
+            for i, k in enumerate(inspect.getfullargspec(f).args)
+            if k not in kwargs.keys() and i < len(args)
```
aguschin, line 46: could it be that `k in kwargs.keys()` if `i < len(args)`? Can't generate an example out of my head, so curious. It should be something like this, but it won't work in Python:
```python
def f(a, b, c):
   pass

a, b, c = 1, 2, 3
f(b, a=a, c=c)
```

gto/git_utils.py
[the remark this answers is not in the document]
francesco086: done

tests/api/data/__init__.py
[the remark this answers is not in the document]
francesco086: You could say that. For each public function in a module that you want to test.

What is fixtures OFC?

tests/git_utils/data/__init__.py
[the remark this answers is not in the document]
francesco086: done!

gto/git_utils.py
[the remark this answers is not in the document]
aguschin: I was thinking about something like [this](https://www.geeksforgeeks.org/context-manager-using-contextmanager-decorator/) to isolate temp dir handling from the wrapper itself:
```python
from contextlib import contextmanager

@contextmanager
def git_clone(repo: str) -> str:
    tmp_dir = TemporaryDirectory()
    logging.debug("clone %s in directory %s", repo, tmp_dir)
    Repo.clone_from(url=repo, to_path=tmp_dir)
    yield tmp_dir
    tmp_dir.cleanup()
```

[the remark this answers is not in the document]
aguschin: Then it would be
```python
def git_clone_if_repo_is_remote(f: Callable):
    ...
            try:
                with git_clone(repo=repo) as tmp_dir:
                    return f(tmp_dir, *args, **kwargs)
            except (NotADirectoryError, PermissionError) as e:
```

gto/git_utils.py
francesco086: tbh I copied paste this piece of code from a small library I wrote some months ago: https://gitlab.com/francesco-calcavecchia/py-os-env-injection
Here I tested extensively all possibilities, including when defaults are set. So I gave it a try (too lazy to think about it again :P ) You are right `i < len(args)` is sufficient. Thanks!

Pushed the change :)

tests/api/data/__init__.py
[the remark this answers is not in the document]
aguschin: LOL sorry
OFC = of course

gto/git_utils.py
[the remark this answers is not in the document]
francesco086: This is basically the suggestion you have already done here: https://github.com/iterative/gto/issues/25#issuecomment-1246585717

But, as said, I strongly advise against using a `yield` statement: https://github.com/iterative/gto/issues/25#issuecomment-1246700348

tests/utils.py
francesco086: change done

tests/api/data/__init__.py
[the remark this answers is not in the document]
francesco086: LOL ok :)

Just let me know what you prefer about the test structure. Perhaps one more idea could be to split inside the folder `tests` into `tests/integration_tests` (the pre-existent) and `tests/unit_tests` (the one I wrote).

I am good with anything, even deleting all my tests :)

gto/git_utils.py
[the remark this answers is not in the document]
aguschin: Hmm, maybe I misunderstand something? This is different, since in fact you just write a context manager and work with that context manager as you should. Note `with git_clone(repo=repo) as tmp_dir` - when `f` function runs and returns the result, the control will be given to the context manager and `tmp_dir.cleanup()` will be called.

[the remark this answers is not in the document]
francesco086: sorry, my bad! Too much multi-tasking.

Ok, will do a change along these lines.

tests/api/data/__init__.py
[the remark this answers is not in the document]
aguschin: Ok, I see the value in splitting tests in multiple files like `test_show.py`, `test_annotate.py`, but since this will not work not for all files (e.g. `test_showcase.py` is clearly an integration test, `test_cli.py` have huge `test_commands` function that makes things faster repeating different read-only commands on a single repo, etc), I'm not sure we should do that right away. Since the project is quite small and there are not so many tests, let's stick to the current structure for now. If you want to re-write tests to fit to what you suggest, let's do it in another PR later.

Also, please rename `data` into `resources` to be consistent [with MLEM ](https://github.com/iterative/mlem/tree/main/tests).

gto/git_utils.py
aguschin: Ok, I see. Let's keep it this then!

gto/git_utils.py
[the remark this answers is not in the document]
francesco086: Done, I introduced `cloned_git_repo`.

Still, I don't want to give up on the decorator able to handle a function where `repo` is not necessarily the first argument :P

gto/git_utils.py
```diff
@@ -15,27 +16,22 @@ def wrapped_f(*args, **kwargs):
...
-        if isinstance(kwargs["repo"], str) and is_url_to_remote_repo(
+        if isinstance(kwargs["repo"], str) and is_url_of_remote_repo(
             repo=kwargs["repo"]
         ):
             try:
-                with TemporaryDirectory() as tmp_dir:
-                    logging.debug("create temporary directory %s", tmp_dir)
-                    # noinspection PyTypeChecker
-                    git_clone(repo=kwargs["repo"], dir=tmp_dir)
+                # noinspection PyTypeChecker
```
aguschin, line 23: Haven't seen this anywhere earlier. This is something to suppress PyCharm suggestions?

gto/git_utils.py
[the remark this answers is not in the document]
aguschin: Yes, let's keep it, no problem)

gto/git_utils.py
```diff
@@ -0,0 +1,71 @@
...
+    logging.debug("create temporary directory %s", tmp_dir)
+    git_clone(repo=repo, dir=tmp_dir.name)
+    yield tmp_dir.name
+    logging.debug("delete temporary directory %s", tmp_dir)
+    tmp_dir.cleanup()
+
+
+def git_clone(repo: str, dir: str) -> None:
+    logging.debug("clone %s in directory %s", repo, dir)
+    Repo.clone_from(url=repo, to_path=dir)
```
aguschin, line 71: Let's merge these functions and test a single one only? Now having `git_clone` as a separate seems redundant. You can easily test `cloned_git_repo` without generating a temporary directory additionally in tests.

tests/api/data/__init__.py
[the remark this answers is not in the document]
francesco086: done

gto/git_utils.py
francesco086: Yes, if it bothers you I can remove it

gto/git_utils.py
francesco086: Can gladly do it for the tests (actually it connects with something I wanted to discuss about public/private functions -> will add it below), but for the functions I would suggest to leave them as they are, again to leave the door open for a different engine. It doesn't do any harm and makes the separation clear: this is the only thing you need to change if you want to use a different git engine.

francesco086:
@aguschin One thing I would like to discuss is that in the module `git_utils` there are now some functions that should probably be private. Currently the only public one (meaning, used outside the module itself) is `git_clone_remote_repo` (the decorator). All others can be turned into private. What do you think?
Maybe one could also argue that `cloned_git_repo` could be public too (perhaps needed in future?).

If you agree, then tests for private functions are probably to be removed to avoid having too rigid tests that could potentially make further development hard.

Function I would like to turn into private and remove the tests of: `git_clone` and `is_url_of_remote_repo`.

Let me know :)

gto/git_utils.py
aguschin: Thanks, I checked out the second PR you created.

My opinion:
1. We should keep tests. We can remove them anytime later. Keeping the functions public is also OK for me, this is in the fashion of the rest of GTO codebase I think.
2. We can keep these functions separate, doesn't matter for now I guess.

I think the PR is ready, no need to overthink this. We can get fix all of that when that would be needed. Let's merge this and move along. WDYT?

gto/git_utils.py
```diff
@@ -0,0 +1,71 @@
...
+
+from gto.constants import remote_git_repo_regex
+
+
+def git_clone_remote_repo(f: Callable):
+    @wraps(f)
+    def wrapped_f(*args, **kwargs):
+        kwargs = _turn_args_into_kwargs(args, kwargs)
+
+        # noinspection PyTypeChecker
```
aguschin, line 18:
```suggestion
```

gto/git_utils.py
aguschin:
```suggestion
```

aguschin (approved):
Two last code suggestions)

gto/git_utils.py
francesco086: fine with me! :) let's do that

francesco086:
@aguschin that should be it... :) whenever if you feel ready to merge

francesco086:
Awesome!! Thank you @aguschin for the patience... aligning on code style is definitely time and energy consuming. But now next features will come very quickly!

PR #273, opened 2022-09-21
Enable remote git repos for `check-ref`, `history`, and `stages`

francesco086:


francesco086:
mmm... why so many commits? I did sync my main with the iterative main before pushing new commits... what should I have done differently?

tests/utils.py
```diff
@@ -35,6 +36,12 @@ def check_obj(
     assert_equals(obj_values, values)
 
 
+def convert_objects_to_str_in_json_serializable_object(
+    o: Union[list, dict]
+) -> Union[list, dict]:
+    return json.loads(json.dumps(o, default=str))
```
aguschin, line 42: Nice! Didn't know there is a `default` arg. Maybe this stuff I wrote earlier wasn't needed with this :) https://github.com/iterative/gto/blob/main/gto/utils.py#L24

aguschin (approved):
Awesome! LGTM!

tests/resources/sample_remote_repo_expected_history_churn.json
```diff
@@ -0,0 +1,93 @@
+[
+  {
+    "timestamp": "2022-08-26 04:55:09",
```
francesco086, line 3: Question here: on my local all timestamp from `gto history` are shown shifted 2 hours forward (e.g. this one is at 6AM).
On GitHub actions I believe they use UTC. So, I changed the timestamp in this json file accordingly, and now the CI is not broken anymore. BUT on my local machine this test is failing, and so on most people's computer, I fear.

I think this should be fixed, and the timestamp should specify the time zone, or refer always to UTC... any idea?

aguschin:
> mmm... why so many commits? I did sync my main with the iterative main before pushing new commits... what should I have done differently?

I assume this is because you didn't fetched upstream in the repo, so after the previous PR was merged, your `main` branch contained all those commits, being branched out from the HEAD of `main` few commits back. NP :) Merging this!

tests/resources/sample_remote_repo_expected_history_churn.json
aguschin: Hmm, let me check on my Mac.

aguschin: Alright, reproduced this. Few options I see so far:
1. Exclude the `timestamp` field from check. Could be done with `for` loop + `tests.utils.check_obj(..., skip_keys=["timestamp"])` I believe. This is what I did before e.g. for `test_registry`. This is also handy since you don't need to change `resources` and fix tests after the test repo is re-generated.
2. Fix datetimes in GTO to be TZ-aware, and then before you `dump` them, make them be in UTC. Will require something like https://github.com/iterative/gto/blob/main/gto/utils.py#L24 I suppose, which will discover datetimes inside a nested objects and serialize them. (Fixing them to be tz-aware is a needed thing on it's own, but only for API consumers like Studio BE, so this is not a direct benefit for this item so far.)

aguschin: 3. Change the timezone for the tests. Looking if this possible. Maybe somehow manupulate shell ENV vars for the `pytest` process, or mock the `datetime` module in Python, dunno yet :)

aguschin: Looks like I found the answer thanks to this [SO answer](https://stackoverflow.com/a/48051938/4890419). Pushed the code, you can check if that works for you. Works for me, local tests pass as expected.

aguschin: I guess I'll need to decide on which approach is better and update the tests to force it.

francesco086: nice, thanks!

what? I thought that frozen means frozen! :D

francesco086:
> 

Got it thanks! Not used to the github fork-pull strategy (I come from GitLab MRs...)

Issue #265, opened 2022-09-13
Write errors to `stderr` instead of `stdout`


aguschin:
After [working on gto-action](https://github.com/iterative/gto-action/pull/7) I realized that's it's inconvenient to handle this in CI. E.g. if you just want to find out the path, you do
```shell
$ export ARTIFACT_PATH=`gto describe mymodel --path`
$ echo $ARTIFACT_PATH
No path exists for artifact
```
which is in fact the error message and should be written to stderr.

The other question is that maybe `gto describe` shouldn't return exitcode 1 here at all. If the description doesn't exist, it's not an incorrect behaviour, incorrect input data or something incorrect at all. Maybe someone wants to check if annotation in `artifacts.yaml` exists at all? Then we may need another command to check it. 

Need to check this for MLEM also.

aguschin:
UPD: confirmed and created an issue for MLEM https://github.com/iterative/mlem/issues/404

aguschin:
Hi @itstargetconfirmed! Wanted to bring to your attention that the GTO project have the same problem as you just solved for MLEM. Do you want to create a PR for that one as well?

itstargetconfirmed:
Hi @aguschin, I'd love to take this on. I'll create a PR tomorrow for this.

aguschin:
closed by #276

PR #276, opened 2022-09-23
Fix/stderr

itstargetconfirmed:
Hi @aguschin, 

I've completed the pull request for https://github.com/iterative/gto/issues/265. 

I've implemented a similar fix for MLEM and further details can be found here: https://github.com/iterative/mlem/pull/410

I've completed the following updates: 
1. Implemented a context manager `stderr_echo` to print to `stderr`.
2. Implemented the printing of `Exception` and `GTOException` to `stderr` as opposed to `stdout` using the `stderr_echo` context manager.
3.  Updated `conftest.py` with a `Runner` class similar to the MLEM project. So that `CliRunner` instances will be created with standard parameters such as `mixed_stderr=False`. With this implementation, we will guarantee that the `result` object will have `stderr` and `stdout` properties. 
4. Implemented a `Runner` pytest fixture, similar to the MLEM project, however, no tests are using it currently. I've added it for future use cases. 
5. Implemented 2 tests that ensure `Exception` and `GTOException` are printed to `stderr`. 
6. Implemented 2 helper functions `_check_output_contains(output: str, search_value: str) -> bool:` and `_check_output_exact_match(output: str, search_value: str) -> bool`. 
7. The functions in item 6 are used to determine how `_check_successful_cmd` and `_check_failing_cmd` should compare `stderr` and `stdout` with their expected output. 
8. The functions in item 6 are needed because in the current implementation, streams are exacted matched with their expected outputs, however, when printing to `stderr` the `gto` application adds the colour red to the output, which prevents exact match. `_check_output_contains` allows us to match a substring, correcting the issue.

### DIFF
--- a/gto/cli.py
+++ b/gto/cli.py
@@ -18,7 +18,16 @@ from gto.constants import (
     VersionSort,
 )
 from gto.exceptions import GTOException, NotImplementedInGTO, WrongArgs
-from gto.ui import EMOJI_FAIL, EMOJI_GTO, EMOJI_OK, bold, cli_echo, color, echo
+from gto.ui import (
+    EMOJI_FAIL,
+    EMOJI_GTO,
+    EMOJI_OK,
+    bold,
+    cli_echo,
+    color,
+    echo,
+    stderr_echo,
+)
 from gto.utils import format_echo, make_ready_to_serialize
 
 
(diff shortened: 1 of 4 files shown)


aguschin (approved):
LGTM! Thanks a lot @itstargetconfirmed!

Issue #272, opened 2022-09-21
Errors in `gto assign` are made for Labels-env approach


aguschin:
Thanks to @omesser to figuring this out. Since `gto assign` behavior in fact depends on the values for `--assignments-per-version` and `--versions-per-stage` options, the absence of them causes confusion. E.g. you expect `Envs-like` approach, which is the default now in `gto show`, but you get `Labels-like` approach for `assign` now.

![image](https://user-images.githubusercontent.com/6797716/191499051-38d8e8b8-9321-4670-a650-a0525b30ead6.png)

![image](https://user-images.githubusercontent.com/6797716/191499119-8c3e5cdc-b29a-449c-9b41-a14536749e1d.png)

aguschin:
UPD: related comment which may cause some implementation adjustment https://github.com/iterative/gto/pull/281/files#discussion_r989914489

Issue #115, opened 2022-04-12
Add `status` command


aguschin:
Would be beneficial to see what git tags are not pushed right now (or possible conflicts for git tags - e.g. if you have local tag and remote but they're different).

E.g.
```
$ gto status
- Two artifacts (mymodel1, mydataset3) are added, but not committed. 
  Use `git add artifacts.yaml && git commit -m "your message"` to commit them.
- One new local git tag that registers a version is detected (`rf@v1.0.1`).
  To push it to the repo use `gto push --tags`
```

francesco086:
I think this ticket is a very good idea! Such command may be of big help.

PR #281, opened 2022-10-03
feat: gto assign / deprecate / register  —auto-push

francesco086:
- close https://github.com/iterative/gto/issues/282
- close https://github.com/iterative/gto/issues/283


francesco086:
@aguschin if you could please give a feedback on this...?

If it is ok, then I would apply the same (add the `--auto-push` option) to all writing commands as part of this PR.

gto/git_utils.py
```diff
@@ -67,3 +67,17 @@ def cloned_git_repo(repo: str):
 def git_clone(repo: str, dir: str) -> None:
     logging.debug("clone %s in directory %s", repo, dir)
     Repo.clone_from(url=repo, to_path=dir)
+
+
+def git_push_tag(repo_path: str, tag_name: str, remote_name: str = "origin") -> None:
+    repo = Repo(path=repo_path)
+    remote = repo.remote(name=remote_name)
+    remote_url = "unknown" if not hasattr(remote, "url") else remote.url
+    logging.debug(
```
aguschin, line 76: I think it would be good if we would be verbose here and communicate how the command changed the world outside of it :)

Like always printing this to stdout:
```shell
$ gto register model --auto-push
Created Git tag `model@v0.0.1`
Running `git push origin model@v0.0.1`
Successfully pushed Git tag(s) to the remote.
```

Or even just printing the real command output (I'm sure `GitPython` calls `git` CLI underneath, so maybe we can just get that output and print it). In this case it would behave just as if you run that command, which is well recognizeable and familiar.

aguschin (approved):
LGTM! Like your mocking tests!

gto/git_utils.py
```diff
@@ -67,3 +67,17 @@ def cloned_git_repo(repo: str):
...
+        "push tag %s from directory %s to remote %s with url %s",
+        tag_name,
+        repo_path,
+        remote_name,
+        remote_url,
+    )
+    remote.push(tag_name)
```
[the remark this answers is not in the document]
francesco086: that's what I would guess too, why?

francesco086:
While I wait for your feedback regarding the echoed message, I extend the same option (`--auto-push`) to all the other writing commands

gto/git_utils.py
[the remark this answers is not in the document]
aguschin: Sorry, didn't get the question :)
I assume this would fail cause you invoke `push`, but `remote_url` is None, so it don't know where to push to. Am I wrong?

gto/registry.py
```diff
@@ -554,3 +559,17 @@ def get_stages(self, allowed: bool = False, used: bool = False):
             return self._get_allowed_stages()
         # if stages aren't set in config, return those in use
         return self._get_used_stages()
+
+    def _push_tag_or_echo_reminder(
+        self, tag_name: str, auto_push: bool, stdout: bool
+    ) -> None:
+        if auto_push:
+            if stdout:
+                echo(f"Running git push origin {tag_name}")
```
aguschin, line 568:
```suggestion
                echo(f"Running `git push origin {tag_name}`")
```

gto/git_utils.py
aguschin: Yes, looks good!

gto/git_utils.py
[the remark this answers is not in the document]
francesco086: `remote_url` is just used in the logging (just a safe wrapper around `remote.url`), I don't think it has nothing to do with failure of `remote.push`.

But I think you are onto something...?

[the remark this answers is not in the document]
aguschin: ok, I was trying to check if there are cases when you actually get a remote without URL (`remote_url == "unknown"` eventually), in this case `remote.push` could fail (because there is no actual remote to use). In this case it would be good to throw some GTO error (like `WrongArg` for example), and not a GitPython one.

francesco086:
I covered all cli writing commands except for:
- `annotate`
- `remove`

tests/test_api.py
```diff
@@ -354,3 +356,127 @@ def test_if_describe_on_remote_git_repo_then_return_expected_info():
...
+    )
+
+
+def test_if_unassign_with_auto_push_then_invoke_git_push_tag(repo_with_artifact):
+    repo, _ = repo_with_artifact
+    gto.api.assign(
+        repo.working_dir, name="model", stage="dev", ref="HEAD", auto_push=False
+    )
+    with patch("gto.registry.git_push_tag") as mocked_git_push_tags:
+        # TODO: ask alex why this does not work without specifying ref="HEAD"
```
francesco086, line 393: Before merging: could you please help me here?

francesco086: forgot to cite @aguschin

gto/registry.py
```diff
@@ -466,17 +484,22 @@ def _return_event(self, tag) -> TBaseEvent:
     @staticmethod
     def _echo_git_suggestion(tag):
         echo("To push the changes upstream, run:")
-        echo(f"    git push {tag}")
+        echo(f"    git push origin {tag}")
 
-    def _delete_tags(self, tags, stdout):
+    def _delete_tags(self, tags, stdout, auto_push: bool):
         tags = list(tags)
+        print()
```
aguschin, line 491: You have these prints across the code. Is this the debug output you forgot to remove?

francesco086: yes :/ I should really find out why the debugger does not work. I removed them now, thanks

tests/test_api.py
aguschin: Next lines are
```python
        gto.api.unassign(
            repo.working_dir, name="model", stage="dev", ref="HEAD", auto_push=True
        )
```
You unassign a `stage` from a `model version`. Generally, you need to specify the version OR `ref`, that pins down the right version. With `Envs-like` approach we use now, you could remove this limitation, since there is only one model version in a stage at particular moment. But, GTO also tries to support `Kanban-like` approach and `Label-like` approach - [docs](https://github.com/iterative/gto#enabling-multiple-versions-in-the-same-stage-workflow). In those cases you need to keep this limitation in place. That's why it's here. Also, for the reference, we have an [issue somehow related to this](https://github.com/iterative/gto/issues/272).

francesco086: thanks for the clarification, I removed the comment from the code

francesco086:
@aguschin I think it's ready to merge?

gto/cli.py
```diff
@@ -562,6 +566,9 @@ def assign(
     message: Optional[str] = option_message,
     simple: str = option_simple,
     force: bool = option_force,
+    auto_push: bool = Option(
```
aguschin, line 569: Looks like `assign --auto-push` doesn't push the first Git tag that is created to register a version:
```bash
$ gto promote ohmypromotion --stage dev --auto-push
Created git tag 'ohmypromotion@v0.0.1' that registers version
To push the changes upstream, run:
    git push origin ohmypromotion@v0.0.1
Created git tag 'ohmypromotion#dev#1' that assigns stage to version 'v0.0.1'
Running `git push origin ohmypromotion#dev#1`
Successfully pushed git tag ohmypromotion#dev#1 to the remote.
```

aguschin: Open question: is it better to do it like this (create tag, push tag, create tag, push tag), or do it together (create both tags, push both)?

gto/cli.py
```diff
@@ -619,6 +627,9 @@ def deprecate(
     simple: str = option_simple,
     force: bool = option_force,
     delete: bool = option_delete,
+    auto_push: bool = Option(
```
aguschin, line 630: A bug 🤔 
```shell
$ gto register bug --auto-push
Created git tag 'bug@v0.0.1' that registers version
Running `git push origin bug@v0.0.1`
Successfully pushed git tag bug@v0.0.1 to the remote.

$ gto deprecate bug --auto-push --delete
Deleted git tag 'bug@v0.0.1'
Running `git push origin bug@v0.0.1`
❌ Unexpected error: 'str' object has no attribute 'new_message_handler'
Please report it here running with '--traceback' flag: <https://github.com/iterative/gto/issues>
```

gto/git_utils.py
```diff
@@ -67,3 +68,26 @@ def cloned_git_repo(repo: str):
 def git_clone(repo: str, dir: str) -> None:
     logging.debug("clone %s in directory %s", repo, dir)
     Repo.clone_from(url=repo, to_path=dir)
+
```
aguschin, line 71: Something to consider: it could be that we have some Git tags locally, but don't have them in remote. Then if you run `--auto-push` in locally cloned repo, you'll decide what to do based on the local repo, but you will try to apply it to the remote one. It could be an error in some cases (you try to remove a remote Git tag that doesn't exist in remote). It could be a incorrect decision (you deprecate a version that doesn't exist in the remote). Looks too complicated to resolve at Friday evening 🍻

aguschin:
Overall looks good, but found couple of bugs @francesco086!

gto/cli.py
francesco086: argh! well spotted... My expectations for the test was wrong! I think I fixed it with this commit https://github.com/iterative/gto/pull/281/commits/ea2df8bfb63d7311b6d481cddb560fe23e7acbd3

gto/git_utils.py
francesco086: "From great power comes great responsibility" I'd say :D
In the end the auto-push is something you can choose to use or not at every call, as you can choose or not to git push a tag. But if this were not (very) useful, why would you echo a suggestion to push the created tags?

In practice I find hard to imagine to work with a local repo out of sync with the remote. This is the CI/CD practice, isn't it?
In fact I would suggest to be able to set auto-push always true in a config file...

gto/cli.py
aguschin: For some reason it doesn't fail if `--stage` arg is not provided, causing this:
```shell
$ gto assign bugfixed --auto-push
Created git tag 'bugfixed@v0.0.1' that registers version
Running `git push origin bugfixed@v0.0.1`
Successfully pushed git tag bugfixed@v0.0.1 to the remote.
Created git tag 'bugfixed#None#1' that assigns stage to version 'v0.0.1'
Running `git push origin bugfixed#None#1`
Successfully pushed git tag bugfixed#None#1 to the remote.
❌ Cannot parse tag name 'bugfixed#None#1'
```

gto/git_utils.py
aguschin: > In practice I find hard to imagine to work with a local repo out of sync with the remote.

This could happen by chance. But ok, I guess we'll need to accommodate for this in the future. This may happen as part of implementing `gto status` command, so I'll write it [in that ticket](https://github.com/iterative/gto/issues/115#issue-1201663040).

> In fact I would suggest to be able to set auto-push always true in a config file...

Good idea!

gto/cli.py
francesco086: mmm... I can't see how it can be related to what I have done in this PR. Behavior is same without the `--auto-push`, can you confirm? This is something that probably should be fixed in a separate PR?

Perhaps now is more urgent because one may push invalid tags (like `bugfixed#None#1`)?

francesco086:
> Overall looks good, but found couple of bugs @francesco086!

Thanks a lot @aguschin , I think I solved them :) except [this](https://github.com/iterative/gto/pull/281#discussion_r990294060), but I don't it's related to this PR - please correct me if I am wrong

gto/cli.py
aguschin: Yes, it doesn't depend on `--auto-push`. We can do it in a separate PR for sure. Created a separate issue for that https://github.com/iterative/gto/issues/285

gto/git_utils.py
aguschin: Example of the currently misleading behavior. Consider you have a git tag `bug@v0.0.1` in remote, but don't have it locally. Then:

```
$ gto register bug --auto-push
Created git tag 'bug@v0.0.1' that registers version
Running `git push origin bug@v0.0.1`
WARNING:git.remote:Error lines received while fetching: error: failed to push some refs to 'https://github.com/aguschin/example-gto'
Successfully pushed git tag bug@v0.0.1 on remote.
```

I think we should fail at push, printing something like: `The Git tag bug@v0.0.1 already exists on remote.` WDYT?

gto/cli.py
```diff
@@ -517,6 +517,9 @@ def register(
     bump_patch: bool = Option(
         False, "--bump-patch", is_flag=True, help="Bump patch version"
     ),
+    auto_push: bool = Option(
+        False, "--auto-push", is_flag=True, help="Push created tag automatically"
```
aguschin, line 521: Let's mark it with `experimental` for now? I anticipate some problems when remote diverges from local, so would be good to warn users we're testing and developing this functionality.
```suggestion
        False, "--auto-push", is_flag=True, help="Push created tag automatically (experimental)"
```

francesco086: good idea! I implemented the change in this https://github.com/iterative/gto/pull/281/commits/a2b0b30c0ef24dedd5cf85350186100faa3f8a65 (the change had to be done in 3 different places)

gto/git_utils.py
francesco086: I agree that it would be more user friendly, therefore desirable. But on the other hand, the same exact error would happen if the user does the git push manually, right? So, as the user can choose if to do git push or not, can also choose whether to add the option `--auto-push` or not.

In conclusion, I don't see much harm in it, but one could make error messages easier to interpret, I agree.

francesco086: Shall we create a new issue for this?

aguschin: Ok, think I got your point. As I see the difference
1. We're communicating it incorrectly. We say `Successfully pushed git tag...`. E.g. the user may assume the CI is triggered in the remote. But the tag was not pushed, and CI was not triggered. So we're confusing user with this message.
2. Often, a user registers a version or assigns a stage to trigger CI. In this case the goal wasn't achieved, so we should exit with code 1. It could be that user don't want any CI and just want the remote repo to be in this state (so as the Git tag was pushed, we don't need to do anything and it is fine). In this case it may be ok, but I think triggering the CI is more frequent goal here. We may add some tag (`--no-fail-if-tags-exist-in-remote` e.g.), but I think it's overkill now.

francesco086: ok, then is it ok if I apply the following change: if `git push` fails, then echo an appropriate message and exit with code 1?

francesco086: appropriate message being something like "Execution of `git push origin bug@v0.0.1` failed. Make sure your local repository is in sync with the remote."

ok?

aguschin: Yes, totally. Thanks 🙏🏻

francesco086: This should be it: https://github.com/iterative/gto/pull/281/commits/0f39722686ab42f29a3d60de37f904089e013975

francesco086:
Please wait before merging (in case everything is fine), I would like to clean up a little bit the tests

francesco086:
> Please wait before merging (in case everything is fine), I would like to clean up a little bit the tests

finished! https://github.com/iterative/gto/pull/281/commits/3858a3f7588b7fd92e5e3c4d3ca7753029b8c96d

aguschin:
Great! It took some time to fix different moments, but I'm 100% sure it was worth it) Thanks a lot for the implementation!

PR #286, opened 2022-10-11
feat: gto register / assign / deprecate support for remote repo

francesco086:


gto/git_utils.py
```diff
@@ -14,7 +14,7 @@
 def git_clone_remote_repo(f: Callable):
```
francesco086, line 14: @aguschin I was thinking to rename this as `git_clone_on_remote_repo`. What do you think?

gto/git_utils.py
```diff
@@ -32,16 +32,21 @@ def wrapped_f(*args, **kwargs):
...
-            k: args[i]
-            for i, k in enumerate(inspect.getfullargspec(f).args)
-            if i < len(args)
-        }
-        kwargs.update(kwargs_complement)
-        return kwargs
+    return wrapped_f
+
+
+def auto_push_on_remote_repo(f: Callable):
```
francesco086, line 38: @aguschin I was thinking to rename this as `git_clone_and_auto_push_on_remote_repo`. What do you think?

gto/git_utils.py
aguschin: Doesn't matter much for me TBH, anything will work. Maybe something like `git_clone_if_needed`?

gto/git_utils.py
aguschin: I think `auto_push` is good since you have an argument that you're setting to `True` with this wrapper.

gto/api.py
```diff
@@ -87,6 +87,7 @@ def remove(repo: Union[str, Repo], name: str):
     return init_index_manager(path=repo).remove(name)
 
 
+@auto_push_on_remote_repo
```
aguschin, line 90: Not sure what caused this, but
```
$ gto register lol -r https://github.com/aguschin/example-gto
Created git tag 'lol@v0.0.1' that registers version
Running `git push origin lol@v0.0.1`
❌ The command `git push origin lol@v0.0.1` failed. Make sure your local repository is in sync with the remote.
```

gto/git_utils.py
francesco086: If I would see a decorator named `git_clone_if_needed` I would wonder "when could it be needed? Perhaps if a repo is remote? Perhaps if one does not want to mess the current repo and prefer to do a dry-run? Perhaps... why didn't they specify it? It must be something complex!"

I think it is better to be specific whenever possible. `git_clone_if_repo_is_remote` could also work, but it makes the test names really hard to read (e.g. `test_git_clone_remote_repo_if_repo_is_a_meaningless_string_then_leave_it_unchanged` would become `test_git_clone_if_repo_is_remote_if_repo_is_a_meaningless_string_then_leave_it_unchanged`).
This is why I was looking for an alternative to a `if` preposition and I came up with the idea of `on`.

gto/api.py
francesco086: Mmm :( this is what happens when you use mocks so much and you don't know enough the library you are mocking.

Anyway, I pushed a fix on how I use gitpython https://github.com/iterative/gto/pull/286/commits/be4e34325516dd5a12e3c47e65244eeb63acd6d2

To me it seems to solve the issue, please double check

gto/git_utils.py
francesco086: ok, you convinced me :)

gto/git_utils.py
aguschin: Got it. To me both `test_git_clone_remote_repo_if_repo_is_a_meaningless_string_then_leave_it_unchanged` and `become test_git_clone_if_repo_is_remote_if_repo_is_a_meaningless_string_then_leave_it_unchanged` look understandable though very long, but I assume it's normal for tests)

Interesting idea about dry-run, btw. Didn't consider this before.

As the other option, I think even `clone_remote_repo` will be fine. There is no other clone in this context except for `git clone`, so it's shouldn't be ambiguous.

gto/api.py
aguschin: Checked, works! Thanks!

gto/git_utils.py
aguschin: Not merging this, in case you want to rename things. Btw, are you able to merge this PR?

aguschin (approved):
Nice! 🚀

gto/git_utils.py
francesco086: I eventually think renaming as `clone_on_remote_repo` made a lot of sense, also for consistency with the new `auto_push_on_remote_repo`.
https://github.com/iterative/gto/pull/286/commits/4dd46fe3bde14fabed366426915fe1fc8df52ef0

No, I cannot merge, I can only close it.

aguschin:
See https://github.com/iterative/gto/issues/25 for details

PR #290, opened 2022-10-18
Annotate and remove support auto commit

francesco086:

### DIFF
--- a/gto/cli.py
+++ b/gto/cli.py
@@ -368,7 +368,7 @@ option_auto_push = Option(
 
 option_auto_commit = Option(
     False,
-    "--auto-push",
+    "--auto-commit",
     is_flag=True,
     help="Automatically commit changes due to this command (experimental)",
 )
--- a/gto/cli.py
+++ b/gto/cli.py
@@ -503,13 +503,17 @@ def annotate(
 
 
 @gto_command(section=CommandGroups.enriching)
-def remove(repo: str = option_repo, name: str = arg_name):
+def remove(
+    repo: str = option_repo,
+    name: str = arg_name,
+    auto_commit: bool = option_auto_commit,
+):
     """Remove the enrichment for given artifact
 
     Examples:
          $ gto remove nn
     """
-    gto.api.remove(repo, name)
+    gto.api.remove(repo, name, auto_commit)
 
 
 @gto_command(section=CommandGroups.modifying)
--- a/gto/cli.py
+++ b/gto/cli.py
@@ -366,6 +366,13 @@ option_auto_push = Option(
     help="Push created tag automatically (experimental)",
 )
 
+option_auto_commit = Option(
+    False,
+    "--auto-push",
+    is_flag=True,
+    help="Automatically commit changes due to this command (experimental)",
+)
+
 
 @app.callback("gto", invoke_without_command=True, no_args_is_help=True)
 def gto_callback(
@@ -472,6 +479,7 @@ def annotate(
     ),
     label: List[str] = Option(None, "--label", help="Labels to add to artifact"),
     description: str = Option("", "-d", "--description", help="Artifact description"),
+    auto_commit: bool = option_auto_commit,
     # update: bool = Option(
     #     False, "-u", "--update", is_flag=True, help="Update artifact if it exists"
     # ),
@@ -489,6 +497,7 @@ def annotate(
         must_exist=must_exist,
         labels=label,
         description=description,
+        auto_commit=auto_commit,
         # update=update,
     )
 
--- a/gto/commit_message_generator.py
+++ b/gto/commit_message_generator.py
@@ -1,0 +1,19 @@
+from typing import Optional
+
+
+def generate_annotate_commit_message(
+    name: str, type: Optional[str] = None, path: Optional[str] = None
+) -> str:
+    return (
+        f"Annotate artifact {name}"
+        f"{f' of type {type}' if type is not None else ''}"
+        f"{f' with path {path}' if path is not None else ''}"
+    )
+
+
+def generate_remove_commit_message(name: str) -> str:
+    return f"Remove annotation for artifact {name}"
+
+
+def generate_empty_commit_message() -> str:
+    return ""
--- a/gto/commit_message_generator.py
+++ b/gto/commit_message_generator.py
@@ -5,9 +5,9 @@ def generate_annotate_commit_message(
     name: str, type: Optional[str] = None, path: Optional[str] = None
 ) -> str:
     return (
-        f"Annotate artifact {name}"
-        f"{f' of type {type}' if type is not None else ''}"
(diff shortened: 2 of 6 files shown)


francesco086:
There are two things missing yet:

1. (that I would like to discuss) what should be the commit message? I was thinking to use the command used, something like `gto.api.annotate(repo=., name="new-type", auto_commit=True)`
2. raise a gto exception if are trying to commit the changes on a file different from HEAD (as suggested in https://github.com/iterative/gto/issues/25#issuecomment-1277207211)

gto/cli.py
```diff
@@ -472,6 +479,7 @@ def annotate(
     ),
     label: List[str] = Option(None, "--label", help="Labels to add to artifact"),
     description: str = Option("", "-d", "--description", help="Artifact description"),
+    auto_commit: bool = option_auto_commit,
```
aguschin, line 482: WDYT about using just `commit` instead of `auto_commit`? For CLI and API. And `push` instead of `auto_push`. Looks like it's sufficient and shorter.

gto/cli.py
```diff
@@ -366,6 +366,13 @@ def callback_sort(  # pylint: disable=inconsistent-return-statements
     help="Push created tag automatically (experimental)",
 )
 
+option_auto_commit = Option(
+    False,
+    "--auto-push",
```
aguschin, line 371: `--auto-commit` I guess?

gto/cli.py
```diff
@@ -489,18 +497,23 @@ def annotate(
...
         # update=update,
     )
 
 
 @gto_command(section=CommandGroups.enriching)
-def remove(repo: str = option_repo, name: str = arg_name):
+def remove(
+    repo: str = option_repo,
+    name: str = arg_name,
+    auto_commit: bool = option_auto_commit,
```
aguschin, line 509: > what should be the commit message? I was thinking to use the command used, something like `gto.api.annotate(repo=., name="new-type", auto_commit=True)`

To your first question. Sounds ok to me. It can also be something like "Annotated artifact 'new-type' with path='path/to/it', type='type-of-it', ...".

As a small side quest we could also add `--message` arg (like there is for `gto register`), to allow writing custom commit messages.

aguschin (commented):
Looks good!

gto/cli.py
francesco086: Very good idea!

However, may I suggest for this PR to stick to `auto_commit`, for consistency with the pre-existing `auto_push`?
After this PR is merged, I will create a separate PR to rename all of them :)

I don't like mixed up PRs, as you have may noticed :P

gto/cli.py
francesco086: Thanks! fixed in https://github.com/iterative/gto/pull/290/commits/66c5728bb15be0d42b4dba49df8a9ff95fa129e1

gto/cli.py
francesco086: Good, will do that then.

Regarding the `--message` option though I am not 100% convinced: it adds complexity to a CLI which is already not that straightforward, and I wonder if it will ever be needed. If it is really needed in special circumstances, one can still commit manually with a git command.
I'd suggest to leave it out for now, and think of it only if some user request it. Do you agree?

aguschin: Yep, totally agree. Let's keep it out of the scope for now.

francesco086:
Hi @aguschin , I think it's ready, I covered the two points left.

Please have a look, in couple of points I had the impression that perhaps the code was getting difficult to read.
In fact I even added a docstring for `gto.git-utils.commit_produced_changes_on_auto_commit`, something I try to avoid as a general rule.
Looking forward to hear your feedback!

Oh, and a small thing: could you please assign this PR to me? I am not clear if this is important for the Hacktoberfest, but just in case... :P

gto/commit_message_generator.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+def generate_annotate_commit_message(
+    name: str, type: Optional[str] = None, path: Optional[str] = None
+) -> str:
+    return (
+        f"Annotate artifact {name}"
+        f"{f' of type {type}' if type is not None else ''}"
+        f"{f' with path {path}' if path is not None else ''}"
+    )
```
aguschin, line 11:
```suggestion
    return (
        f"Annotate artifact `{name}`"
        f"{f' of type `{type}`' if type is not None else ''}"
        f"{f' with path `{path}`' if path is not None else ''}"
    )
```

gto/git_utils.py
```diff
@@ -51,6 +52,80 @@ def wrapped_f(*args, **kwargs):
...
+                        result = f(**kwargs)
+                        if are_files_in_repo_changed(
+                            repo_path=kwargs["repo"],
+                            files=stashed_tracked + stashed_untracked,
+                        ):
+                            _reset_repo_to_head(repo_path=kwargs["repo"])
+                            raise GTOException(
+                                msg="The command changed files that were not committed, automated committing is not possible. "
+                                "The files have been rolled back to their previous state.\n\n"
+                                "Suggested action: Commit the changes and re-run this command"
```
aguschin, line 97:
```suggestion
                                msg="The command changed files that were not committed, automated committing is not possible.\n"
                                "Suggested action: Commit the changes and re-run this command"
```
"...have been rolled back..." - user doesn't know we did stash, so no need to tell this to him.

gto/git_utils.py
```diff
@@ -51,6 +52,80 @@ def wrapped_f(*args, **kwargs):
...
+                        git_add_and_commit_all_changes(
+                            repo_path=kwargs["repo"],
+                            message=generate_commit_message(**kwargs),
+                        )
+                else:
+                    raise ValueError(
+                        "Function decorated with commit_produced_changes_on_auto_commit was called with "
+                        "`auto_commit=True` but `repo` was not provided."
+                        "Argument `repo` is necessary."
+                    )
```
aguschin, line 108: User won't see this, only the developer may see it - when he wraps a function without `repo` arg, right?

aguschin (approved):

gto/git_utils.py
francesco086: correct!

gto/git_utils.py
francesco086: What about `The command would have changed files ...` instead of `The command changed files ...`.

I want to make clear to the user that the command did not have any effect and everything is as before.

aguschin: Yes, agreed! Missed that line. Fixed the suggestion above. Looks good now?

francesco086: Changed it in https://github.com/iterative/gto/pull/290/commits/666e7edba0ffaff4b6e10fdb0abb5a0686ff3661

aguschin:
Awesome! Merging this. Thanks @francesco086!

PR #291, opened 2022-10-20
set black line-length to 120 to improve readibility

francesco086:
Hi @aguschin , may I suggest to increase the line length to 120 (set in `pyproject.toml`)?
It makes the code way more readable imho. What do you think?

### DIFF
--- a/gto/api.py
+++ b/gto/api.py
@@ -51,9 +51,7 @@ def _is_gto_repo(repo: Union[str, Repo]):
 def _get_index(repo: Union[str, Repo], file=False):
     """Get index state"""
     if file:
-        return FileIndexManager.from_path(
-            path=repo if isinstance(repo, str) else repo.working_dir
-        )
+        return FileIndexManager.from_path(path=repo if isinstance(repo, str) else repo.working_dir)
     return RepoIndexManager.from_repo(repo)
 
 
@@ -68,9 +66,7 @@ def get_stages(repo: Union[str, Repo], allowed: bool = False, used: bool = False
 
 
 # TODO: make this work the same as CLI version
-@commit_produced_changes_on_auto_commit(
-    message_generator=generate_annotate_commit_message
-)
+@commit_produced_changes_on_auto_commit(message_generator=generate_annotate_commit_message)
 def annotate(
     repo: Union[str, Repo],
     name: str,
@@ -94,12 +90,8 @@ def annotate(
     )
 
 
-@commit_produced_changes_on_auto_commit(
-    message_generator=generate_remove_commit_message
-)
-def remove(
-    repo: Union[str, Repo], name: str, auto_commit: bool = False
-):  # pylint: disable=unused-argument
+@commit_produced_changes_on_auto_commit(message_generator=generate_remove_commit_message)
+def remove(repo: Union[str, Repo], name: str, auto_commit: bool = False):  # pylint: disable=unused-argument
     """Remove an artifact from the Index"""
     return init_index_manager(path=repo).remove(name)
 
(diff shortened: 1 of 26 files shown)


aguschin:
Hi @francesco086! Mixed feelings: in some places it's more readable (where we have long function names), in others less readable (where we have short var names > a lot of things happen in a single line). Let's keep it as it is :)

Btw, will you create a PR to rename `--auto-push` and `--auto-commit`? This may shorten some function names as well I assume :)

skshetry:
The problem here seems to be unnecessary verbosity. I suggest adopting something like follows during PR reviews (`Studio` also has a similar guidelines if you check it's `BE review cheatsheet`):

> Identifiers that exist for short scopes should be short,
> whereas identifiers that live for large scopes should be larger and more descriptive.

Although in the second statement, I'll still suggest cutting down verbosity from larger scopes. You can use context from surrounding context from `namespaces`/`modules`/`class`, etc.

Eg: `are_files_in_repo_changed` could be just `files_changed()` because they are inside `git_utils` namespace. `clone` on remote repo could be just `clone()`. `@commit_produced_changes_on_auto_commit` could be `@auto_commit` (on produced changes is implied), etc.

francesco086:
Hi @skshetry , thanks for the hints. The idea that names for small contexts can be small, whereas for large scope should be long is in general a good one, it's one of the suggestion in Clean Code.

However I find `files_changed()` a bad name: it breaks the "rule" (again, from Clean Code) that function names should be verbs.
`clone` alone is not descriptive enough imho, this is a public function. How can someone understands what is doing it without checking the code?
I strongly believe that when in doubt a little extra verbosity does not harm.

@aguschin I'm fine with the decision, but I would like to mention that breaking the line-length to 72 assumes that someone is checking the code on a window that small, which nowadays is pretty much never true :)
Do we still live in the time of [FORTRAN](https://en.wikipedia.org/wiki/Fortran)? :D

Anyhow, in the end is your package, so it's up to you.

PS: @aguschin yes, I am going to rename `auto-commit` and `auto-push` for sure. But first `--auto-push` for `annotate` and `remove` (I already started), and support for remote repos. It won't take long.

skshetry:
> it breaks the "rule"

It does not have side effects, so it's fine.
If you need to follow that rule, you can name it `have/are_files_changed()`. 
(Again, I don't think we follow that rule in Python, `sorted()` builtin comes to mind.)

> this is a public function

only `gto.api` is public AFAIK. 

> How can someone understands what is doing it without checking the code?

That 's the thing, they don't have to, they should depend on the abstraction. The `clone_on_remote_repo` is doing dependency injection in an unconventional way, maybe that is the problem. Instead of taking a callable, maybe directly using it would be more readable. In dvc, we have `Repo.open()` that can clone external repo conditionally, and returns you a `Repo` instance.

> I strongly believe that when in doubt a little extra verbosity does not harm.

+1 to that, but if it isn't adding anything, it is better to keep things simple and concise.

> All the great things are simple, and many can be expressed in a single word ...

My comment is more general to the codebase than just the picked examples. At least, for my taste, It is too verbose.

francesco086:
@skshetry I agree with you on most things, but what you are especially right about, is that the decorators I created feel very unconventional to me as well. Perhaps this is the source of problem! It is very, very much a functional programming style, and at least I am not used to it. So, without rather verbose names, I would be totally confused myself. And I would like avoid falling in the scenario: "When I wrote the code, only God and I knew what it was doing. Now, only God does." XD

The approach of the decorators (and therefore functional style) was agreed with @aguschin at the beginning, for valid technical reasons. I do think is a good idea, I like it, and it was a good opportunity to learn too.

But names and structure are getting more and more messy imo. Right now I am focused on finished the functionalities, but I would like to clean it up for good at the end. Do you mind if I involve you (@skshetry)  in the renaming and refactoring part? (I take @aguschin  for granted :P)

In conclusion, please @aguschin bear with me for a couple of more PRs, then I intend to push major clean up PRs.
If we do it now I am afraid we will get lost in a lot of discussions and don't make progress on the functionalities. And we are almost there!

aguschin:
> When I wrote the code, only God and I knew what it was doing. Now, only God does

😂 

Thanks for the feedback @skshetry. Your advice here is very useful.

@francesco086, sure, let's finish with the PRs, sure. Besides, AFAIK @skshetry has a vacation next week.

aguschin:
@skshetry, please TAL at this PR by @francesco086 https://github.com/iterative/gto/pull/296

PR #292, opened 2022-10-21
gto annotate and remove support auto push and remote repo

francesco086:
Contribute to https://github.com/iterative/gto/issues/25

### DIFF
--- a/gto/git_utils.py
+++ b/gto/git_utils.py
@@ -124,7 +124,14 @@ def push_on_auto_push(f: Callable):
             kwargs["auto_commit"] = True
             result = f(**kwargs)
             if "repo" in kwargs:
-                git_push(repo_path=kwargs["repo"])
+                try:
+                    git_push(repo_path=kwargs["repo"])
+                except Exception as e:
+                    raise GTOException(  # pylint: disable=raise-missing-from
+                        "It was not possible to run `git push`. "
+                        "The detailed error message was:\n"
+                        f"{str(e)}"
+                    )
             else:
                 raise ValueError(
                     "Function decorated with push_on_auto_push was called with "
--- a/gto/git_utils.py
+++ b/gto/git_utils.py
@@ -70,7 +70,8 @@ def commit_produced_changes_on_auto_commit(
 
     def generate_commit_message(**kwargs) -> str:
         kwargs_for_message_generator = {
-            k: kwargs[k] for k in inspect.getfullargspec(message_generator).args
+            k: kwargs.get(k, None)
+            for k in inspect.getfullargspec(message_generator).args
         }
         return message_generator(**kwargs_for_message_generator)
 
--- a/gto/git_utils.py
+++ b/gto/git_utils.py
@@ -36,7 +36,7 @@ def clone_on_remote_repo(f: Callable):
     return wrapped_f
 
 
-def auto_push_on_remote_repo(f: Callable):
+def set_auto_push_on_remote_repo(f: Callable):
     @wraps(f)
     def wrapped_f(*args, **kwargs):
         kwargs = _turn_args_into_kwargs(f, args, kwargs)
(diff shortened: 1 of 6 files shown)


francesco086:
I apologize upfront for some terrible names like `push_on_auto_push`... but functionalities should be there.

Will go on on Monday

francesco086:
TODO:  catch the error in case `git push` fails and display a user friendly message.

gto/git_utils.py
```diff
@@ -116,6 +116,28 @@ def wrapped_f(*args, **kwargs):
     return wrap
 
 
+def push_on_auto_push(f: Callable):
```
aguschin, line 119: I see that we use different approaches to push Git references for `assign` (and others) and `annotate` (and others). Could we maybe unite them? E.g. the API method should return Event(s), and wrapper should get Git references from those events, and push? E.g. the same wrapper would invoke `git push origin abc1234` after annotate, and `git push origin nn@v0.0.1` after `register`. WDYT?

aguschin: This approach would also do `push` only after the whole gto command succeeds. E.g. if you run `assign` and for some reason, the version was registered, but stage wasn't assigned, it won't run `git push` and let you fix things locally before impacting the remote.

aguschin: This may require some other changes though, since the `_return_event` that's invoked in the end of each API call can't currently handle the Git tags we remove. Only the created ones, and I believe if in `assign` we create 2 tags (registration + assignment), it will return the second only.

francesco086: So, this would require changing the interface of the commands in `api.py`, correct?

aguschin: Yes, I believe, what is `return`ed may require a change.

francesco086:
@aguschin do you have any clue about the errors happening for windows? I am quite clueless tbh...

francesco086:
> @aguschin do you have any clue about the errors happening for windows? I am quite clueless tbh...

Issue with windows with py3.8 got solved "on its own" - today the run was successful.

Issue with windows with py3.9 got solved with this fixture: https://github.com/iterative/gto/pull/292/commits/e1432ea5b0c9f21e41c214404d2c7e34cb7901eb (I set it to the default value on my mac)

gto/git_utils.py
aguschin: > check the tags before the command is executed, then after the command is executed, and push only the difference.

Sounds as a good idea, but don't like that if the repo will change in the meantime (while the command runs), that will result in a wrong diff.

aguschin: > Then I would suggest to leave out the decision for another round

Do you suggest handling this in a separate issue then? I guess it could be in the same place as the other refactoring things (renaming functions, possibly using something but the functional approach, etc).

francesco086: Yes, that's what I am suggesting. It's anyway quite a mess, so let's get the thing to work, and then refine the code. This is what we have done until now, and it has the positive side that now we see the full picture. At the beginning it was unclear to me where we would have ended up.

francesco086: > Sounds as a good idea, but don't like that if the repo will change in the meantime (while the command runs), that will result in a wrong diff.

Oh god, if that happens we are screwed anyway, aren't we? (e.g. registering model v0.0.1, but in the meantime that tag is created by another process) :D Ideally one should lock the repo and unlock the repo while the command is running

aguschin: Maybe locking is a good option. Still, unclear to me which approach is better :)

aguschin:
Ok, congrats @francesco086 :) everything is there and working now, although the refactoring will surely help 😅

gto/git_utils.py
francesco086: Let's discuss it further, I am also not 100% sure what is best. Need to think deeper about it, will come to this point, it's an important design decision, thanks for rising it :)

Just, let's do it in a separate PR :)

PR #296, opened 2022-10-26
WIP: Refactor decorators in git utils

francesco086:

### DIFF
--- a/gto/api.py
+++ b/gto/api.py
@@ -24,7 +24,7 @@ from gto.constants import (
 from gto.exceptions import NoRepo, NotImplementedInGTO, WrongArgs
 from gto.ext import EnrichmentInfo
 from gto.git_utils import (
-    clone_on_remote_repo,
+    clone,
     commit_produced_changes_on_commit,
     push_on_push,
     set_push_on_remote_repo,
@@ -63,14 +63,14 @@ def _get_state(repo: Union[str, Repo]):
     return GitRegistry.from_repo(repo).get_state()
 
 
-@clone_on_remote_repo
+@clone(repo_arg="repo")
 def get_stages(repo: Union[str, Repo], allowed: bool = False, used: bool = False):
     return GitRegistry.from_repo(repo).get_stages(allowed=allowed, used=used)
 
 
 # TODO: make this work the same as CLI version
 @set_push_on_remote_repo
-@clone_on_remote_repo
+@clone(repo_arg="repo")
 @push_on_push
 @commit_produced_changes_on_commit(message_generator=generate_annotate_commit_message)
 def annotate(
@@ -98,7 +98,7 @@ def annotate(
 
 
 @set_push_on_remote_repo
-@clone_on_remote_repo
+@clone(repo_arg="repo")
 @push_on_push
 @commit_produced_changes_on_commit(message_generator=generate_remove_commit_message)
 def remove(
@@ -112,7 +112,7 @@ def remove(
 
 
 @set_push_on_remote_repo
-@clone_on_remote_repo
+@clone(repo_arg="repo")
 def register(
     repo: Union[str, Repo],
     name: str,
@@ -148,7 +148,7 @@ def register(
 
 
 @set_push_on_remote_repo
-@clone_on_remote_repo
+@clone(repo_arg="repo")
 def assign(
     repo: Union[str, Repo],
     name: str,
@@ -184,7 +184,7 @@ def assign(
 
 
 @set_push_on_remote_repo
-@clone_on_remote_repo
+@clone(repo_arg="repo")
 def unassign(
     repo: Union[str, Repo],
     name: str,
@@ -217,7 +217,7 @@ def unassign(
 
 
 @set_push_on_remote_repo
-@clone_on_remote_repo
+@clone(repo_arg="repo")
 def deregister(
     repo: Union[str, Repo],
     name: str,
@@ -248,7 +248,7 @@ def deregister(
 
 
 @set_push_on_remote_repo
-@clone_on_remote_repo
+@clone(repo_arg="repo")
 def deprecate(
     repo: Union[str, Repo],
     name: str,
@@ -307,14 +307,14 @@ def find_versions_in_stage(
     )
 
 
-@clone_on_remote_repo
+@clone(repo_arg="repo")
 def check_ref(repo: Union[str, Repo], ref: str):
     """Find out what have been registered/assigned in the provided ref"""
     reg = GitRegistry.from_repo(repo)
     return reg.check_ref(ref)
 
 
-@clone_on_remote_repo
+@clone(repo_arg="repo")
 def show(
     repo: Union[str, Repo],
     name: Optional[str] = None,
@@ -508,7 +508,7 @@ def _show_versions(  # pylint: disable=too-many-locals
     return versions_, "keys"
 
 
-@clone_on_remote_repo
+@clone(repo_arg="repo")
 def describe(
     repo: Union[str, Repo], name: str, rev: str = None
 ) -> List[EnrichmentInfo]:
@@ -523,7 +523,7 @@ def describe(
     raise NotImplementedError
 
 
-@clone_on_remote_repo
+@clone(repo_arg="repo")
 def history(
     repo: Union[str, Repo],
     artifact: str = None,
--- a/gto/api.py
+++ b/gto/api.py
@@ -24,10 +24,9 @@ from gto.constants import (
 from gto.exceptions import NoRepo, NotImplementedInGTO, WrongArgs
 from gto.ext import EnrichmentInfo
 from gto.git_utils import (
-    clone,
     commit_produced_changes_on_commit,
-    is_url_of_remote_repo,
     push_on_push,
+    remote_to_local,
     set_push_on_remote_repo,
 )
 from gto.index import (
@@ -64,14 +63,14 @@ def _get_state(repo: Union[str, Repo]):
     return GitRegistry.from_repo(repo).get_state()
 
 
-@clone(repo_arg="repo", controller=is_url_of_remote_repo, controller_args=["repo"])
+@remote_to_local(repo_arg="repo")
 def get_stages(repo: Union[str, Repo], allowed: bool = False, used: bool = False):
     return GitRegistry.from_repo(repo).get_stages(allowed=allowed, used=used)
 
 
 # TODO: make this work the same as CLI version
 @set_push_on_remote_repo
-@clone(repo_arg="repo", controller=is_url_of_remote_repo, controller_args=["repo"])
+@remote_to_local(repo_arg="repo")
 @push_on_push
 @commit_produced_changes_on_commit(message_generator=generate_annotate_commit_message)
 def annotate(
@@ -99,7 +98,7 @@ def annotate(
 
 
 @set_push_on_remote_repo
-@clone(repo_arg="repo", controller=is_url_of_remote_repo, controller_args=["repo"])
+@remote_to_local(repo_arg="repo")
 @push_on_push
 @commit_produced_changes_on_commit(message_generator=generate_remove_commit_message)
 def remove(
@@ -113,7 +112,7 @@ def remove(
 
 
 @set_push_on_remote_repo
-@clone(repo_arg="repo", controller=is_url_of_remote_repo, controller_args=["repo"])
+@remote_to_local(repo_arg="repo")
 def register(
     repo: Union[str, Repo],
     name: str,
@@ -149,7 +148,7 @@ def register(
 
 
 @set_push_on_remote_repo
-@clone(repo_arg="repo", controller=is_url_of_remote_repo, controller_args=["repo"])
+@remote_to_local(repo_arg="repo")
 def assign(
     repo: Union[str, Repo],
     name: str,
@@ -185,7 +184,7 @@ def assign(
 
 
 @set_push_on_remote_repo
-@clone(repo_arg="repo", controller=is_url_of_remote_repo, controller_args=["repo"])
+@remote_to_local(repo_arg="repo")
 def unassign(
     repo: Union[str, Repo],
     name: str,
@@ -218,7 +217,7 @@ def unassign(
 
 
 @set_push_on_remote_repo
-@clone(repo_arg="repo", controller=is_url_of_remote_repo, controller_args=["repo"])
+@remote_to_local(repo_arg="repo")
 def deregister(
     repo: Union[str, Repo],
     name: str,
@@ -249,7 +248,7 @@ def deregister(
 
 
 @set_push_on_remote_repo
-@clone(repo_arg="repo", controller=is_url_of_remote_repo, controller_args=["repo"])
+@remote_to_local(repo_arg="repo")
 def deprecate(
     repo: Union[str, Repo],
     name: str,
@@ -308,14 +307,14 @@ def find_versions_in_stage(
     )
 
 
-@clone(repo_arg="repo", controller=is_url_of_remote_repo, controller_args=["repo"])
+@remote_to_local(repo_arg="repo")
 def check_ref(repo: Union[str, Repo], ref: str):
     """Find out what have been registered/assigned in the provided ref"""
     reg = GitRegistry.from_repo(repo)
     return reg.check_ref(ref)
 
 
-@clone(repo_arg="repo", controller=is_url_of_remote_repo, controller_args=["repo"])
+@remote_to_local(repo_arg="repo")
 def show(
     repo: Union[str, Repo],
     name: Optional[str] = None,
@@ -509,7 +508,7 @@ def _show_versions(  # pylint: disable=too-many-locals
     return versions_, "keys"
 
 
-@clone(repo_arg="repo", controller=is_url_of_remote_repo, controller_args=["repo"])
+@remote_to_local(repo_arg="repo")
 def describe(
     repo: Union[str, Repo], name: str, rev: str = None
 ) -> List[EnrichmentInfo]:
@@ -524,7 +523,7 @@ def describe(
     raise NotImplementedError
 
 
-@clone(repo_arg="repo", controller=is_url_of_remote_repo, controller_args=["repo"])
+@remote_to_local(repo_arg="repo")
 def history(
     repo: Union[str, Repo],
     artifact: str = None,
--- a/gto/api.py
+++ b/gto/api.py
@@ -26,6 +26,7 @@ from gto.ext import EnrichmentInfo
 from gto.git_utils import (
     clone,
     commit_produced_changes_on_commit,
+    is_url_of_remote_repo,
     push_on_push,
     set_push_on_remote_repo,
 )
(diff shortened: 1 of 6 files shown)


francesco086:
Hi @aguschin , here a first draft of what I am changing, for now just worked on the cloning decorator. I am writing few points to guide you:
- in `git_utils`:
  - changed the signature of the cloning decorator. It needs: a. the repo b. when to clone c. arguments to use to decide to clone or not
  - renamed `@clone_on_remote_repo` -> `@clone`, I think that its arguments make very clear what happens
  - introduced `_check_required_args_are_provided`, that can be re-used in the other decorators
- in `test_git_utils`:
  - having all the tests in a single file is really painful, I tried at least to cluster them in classes. I still think that separate files would have been clearer, but I don't want to refactor all tests
- in `api`:
  - update the decorations
- in `test_api`
  - just check that the decorator is applied to all the commands that needs it to support remote repos

If you think that this change looks good, I can apply the same to all other decorators.

francesco086:
On a second thought, I believe that passing the controller and its arguments to the cloning decorator is too much.
I should have stuck to the original idea. Also, perhaps, `clone_if_remote` instead of `clone` is necessary to make the code readable...? (I ended up including the controller in the arguments to keep the name short, because I still found it not explicit enough).

However, while working it out, I realized that all decorators in the `git_utils` module have the same structure:
1. they use some arguments given to `f`, and they should check if they were indeed provided
2. there is a if condition (what I called controller), if false run `f` as usual, otherwise do something else
Probably I could abstract it away to avoid code repetition. What do you think?

gto/api.py
```diff
@@ -307,14 +308,14 @@ def find_versions_in_stage(
     )
 
 
-@clone_on_remote_repo
+@clone(repo_arg="repo", controller=is_url_of_remote_repo, controller_args=["repo"])
 def check_ref(repo: Union[str, Repo], ref: str):
     """Find out what have been registered/assigned in the provided ref"""
     reg = GitRegistry.from_repo(repo)
     return reg.check_ref(ref)
```
skshetry, line 315: I may be missing something here, but what is preventing us to do something like follows?

```suggestion
@clone(repo_arg="repo", controller=is_url_of_remote_repo, controller_args=["repo"])
def check_ref(repo: Union[str, Repo], ref: str):
    """Find out what have been registered/assigned in the provided ref"""
    with open_repo(repo) as r:
        reg = GitRegistry.from_repo(r)
        return reg.check_ref(ref)
```

aguschin: Don't you mean to remove decorator at all here?

skshetry: Sorry, fixed.

aguschin: We had a long discussion about it in https://github.com/iterative/gto/issues/25. We decided it's better to use decorator instead of writing this. It looked like a good way to do this at that moment. Maybe it's time to revisit that decision.

aguschin: Btw, if we need to also pass params like `--commit` and `--push` (to automatically commit and push commit or tag), could it work like `with open_repo(repo, commit=commit, push=push) as r: ...`?

aguschin:
Thanks for great summary @francesco086. It helped a lot. WDYT about @skshetry suggestion above?

francesco086:
> 

At the beginning I liked very much the idea of the decorator because it is not intrusive (I didn't need to change the functions in api). If I remember right, this was the reason why we decided to go for decorators, we didn't want to add a `with` statement in each gto command.

However, I found quite painful to work with decorators, especially for what concerns the tests. But perhaps it comes from my inexperience with them.

In the end I don't know what is best :D I would let you decide

aguschin:
> we didn't want to add a with statement in each gto command

Yes, this was one of the reasons as I remember it. Besides CLI commands, Studio will have to refactor a lot of code, if using `with` will be the only option to work with local repos. Maybe using `with` for remote repos and keeping an option to don't use `with` for local ones? Not sure will it look good from the GTO internals perspective though.

Ok, let's see what @skshetry have in mind. I'll play around with this in meantime to see how it can be like.

francesco086:
One question/suggestion: what if `open_repo(repo)` (or the corresponding as decorator), would ALWAYS return a gitpython repo?
Meaning, that one can remove `GitRegistry.from_repo(r)`.

Basically it would be a generalized `GitRegistry.from_repo(r)`, including the remote repos.

Contra I see: tight binding with gitpython (which is there already, but this would reinforce it).

skshetry:
> Basically it would be a generalized GitRegistry.from_repo(r), including the remote repos.

That's what I meant by `open_repo`, that's how we use in `dvc` with `Repo.open()` that always gives you a `Repo` instance:
https://github.com/iterative/gto/pull/296#discussion_r1010224208

francesco086:
Hi @aguschin ! I am living with a bad conscience about the code I left behind me.

Any decision on which direction to take? Just let me know what you consider best.

aguschin:
Hi @francesco086! Sorry, simply didn't have time to investigate this. Will take a look soon.

gto/api.py
aguschin: Ok, there are at least few ways to do it:
1. Leave `with open_repo(repo) as r:` for the user, as in DVC (`with dvc.api.open(...)`) https://dvc.org/doc/api-reference/open
```
with gto.api.open_repo(repo) as r:
    return gto.api.check_ref(...)
```
Cons: This will require Studio to rewrite API calls, and requires to add this `with ...` to all CLI commands. Pros: it allows to re-use cloned repo, and removes the repetitive `with gto.api.open_repo(repo) as r` part from API commands code.
2. Do it inside API calls - this is what @skshetry suggested above.
3. Do both. Provide `gto.api.open_repo` to do the 1st - to re-use cloned repo, and do as @skshetry suggested anyway - in case user supplies a remote repo.

aguschin: And I assume that pro for any of these approaches will be tests simplification, but haven't TAL at that yet.

aguschin: Also, this reminds me of this old request from Studio: when they read GTO repos, they have to iterate a lot on different commits. Bc they call GTO API, `GitRegistry` instance gets recreated for each API call, which slows down things a lot. If we could create not `gitpython.Repo` instance, but `GitRegistry`, that could solve the problem.

The problem with that was: `GitRegistry` provides a single entrypoint (of lower level than `gto.api`), but the state is actually kept in `BaseRegistryState` object. And `BaseRegistryState` instance is re-generated every time it's needed for `GitRegistry` to work with. E.g., some operations like `register` or `assign` can change the state, so sometimes `BaseRegistryState` inside `GitRegistry.register` or `GitRegistry.assign` calls. 🤯

aguschin: But probably, that is another issue that doesn't have to be coupled with reading remote repos. WDYT, @francesco086?

aguschin: btw, isn't `dvc.repo.Repo.open` in fact the same as `gto.registry.GitRegistry.from_repo`? They both return `cls(...)`, although `dvc.repo.Repo.open` clones repo if needed.
They're used to expose some abstraction of DVC repo / GTO repo: e.g. compare this usage https://github.com/iterative/dvc/blob/b2a275ee9e865dec6d632aa96ac90eb385fab468/dvc/api/data.py#L21 with this usage https://github.com/iterative/gto/blob/main/gto/api.py#L313 (missing context manager doesn't count).

We even had [a idea (though with a strange code snippet I must admit)](https://github.com/iterative/gto/issues/25#issuecomment-1246585717) to clone remote repo in `from_repo` method - and `dvc.repo.Repo.open` does the same, although delegates it to `dvc.external_repo` module.

skshetry: `dvc.external_repo` is just an internal implementation detail, and is legacy thing (we did not have scmrepo). Will be refactored when https://github.com/iterative/scmrepo/issues/41 is fixed.

aguschin: btw, @skshetry, have you maybe solved similar problems for Studio as I stated above? In our case the problem is that we need to read all Git tags with each API call - and this is time consuming. I understand you don't need to do that, you rather need to expose what's in a specific commit. But, maybe you still did some similar optimization? Looking for some examples, if they exist.

francesco086: One comment, if I understood correctly, actually the code should be:
```python
def check_ref(repo: Union[str, Repo], ref: str):
    """Find out what have been registered/assigned in the provided ref"""
    with open_repo(repo) as reg:
        return reg.check_ref(ref)
```
right?

francesco086: > If we could create not `gitpython.Repo` instance, but `GitRegistry`, that could solve the problem.

Very sorry, I am not able to follow: what is `GitRegistry`?

francesco086: Sorry for the late answer, yesterday was busy, I tried from the phone but it was a bad idea.

-----------------

Thanks for the explanations @aguschin , many things I was unaware of. From what I understood your suggestions lead to change so that in `cli` (not `api`!) we do
```python
with gto.registry.GitRegistry.from_repo(r) as reg:
    return reg.check_ref(ref)
```
and `check_ref` in `api` could be deleted...? which means that Studio developers will have to adjust their code-base in exchange for performance gain.
Is this what you are suggesting? With the following extra considerations:
1. `GitRegistry.from_repo` should support remote repos
2. `GitRegistry.from_repo` should allow the `commit` and `push` options (or, most likely, not, and find another mechanism)

The only warning sign that comes up to my mind, is that one should be careful not to break the single-responsibility principle, and don't mix all these functionalities together. The single responsibilities I see:
- read a gto registry from local
- clone remote repos
- commit changes
- push tags
- push commit
So, nothing against having a class that allows what I understood you were sketching, but let's be careful about keeping the functionalities distinct and re-usable :)

-----------------

> The problem with that was: GitRegistry provides a single entrypoint (of lower level than gto.api), but the state is actually kept in BaseRegistryState object. And BaseRegistryState instance is re-generated every time it's needed for GitRegistry to work with. E.g., some operations like register or assign can change the state, so sometimes BaseRegistryState inside GitRegistry.register or GitRegistry.assign calls. 

This sounds to me more of a general design problem?

-----------------

These are my two pennies, but I think you have a better overview.

Moving into something practical, so that perhaps I can amend what I have done, and leave general `gto` design decisions and changes to another time: I believe that the simple decorators in `git_utils` (forget the horrible names) respect the rule of doing one thing only, and can be combined with any other class or functions you may create in the future to enhance capabilities, while keeping concerns separated. Would it make sense if I finalize this module, and then use it in any way you consider more appropriate? What I mean is: `@clone_if_remote` could be used to decorate the `api.check_ref`, `GitRegistry.from_repo`, or even `cli.check_ref`. What do you think?

As a matter of fact I was even thinking of taking this `git_utils` out of `gto` and make it a stand-alone mini-library that one can re-use in many project to add support to remote repos.

aguschin: > From what I understood your suggestions lead to change so that in cli (not api!) we do
> ```
> with gto.registry.GitRegistry.from_repo(r) as reg:
>    return reg.check_ref(ref)
> ```
That would just what DVC does. We can remove API - for the sake of development simplicity. It basically just duplicates everything from GitRegistry, and adds only `show`, `history` - and those can be migrated to some "Presenter" class? Or maybe even to `BaseRegistryState` 🤔

aguschin: @francesco086, do you think it would make the tests much easier? I didn't have time to dig into your tests or try to rewrite them, so relying on your opinion here.

francesco086: I am sorry but I am lost, what would make tests easier? Option 1, 2, or 3 in https://github.com/iterative/gto/pull/296#discussion_r1016147255? Could you perhaps rephrase? I am not sure I understand what is it that you mean now.

aguschin: Specifically, what you suggested: use `with gto.registry.GitRegistry.from_repo(r) as reg:` instead of decorators.

francesco086: Ok, now I understand, thanks :)

The answer is: it depends on how we are going to add the remote functionalities to `GitRegistry.from_repo(...)`. For what concern the testing difficulties, this change means only to move the target (if we keep the decorators) :D

This would reduce the number of tests for the cloning functionality, because that would happen only in `from_repo()`. All other functionalities (commit, push) will have to be checked on all (or at least, most) of the `GitRegistry` methods. And if we keep using decorators, there will be the same difficulties.

But, to be honest, when I look at the code itself, I very much like the decorator approach, because there is a very neat separation of concern. What I find very cumbersome is to test them, but maybe it's me, I don't know how to do it in the right way. Perhaps this could be formulated in this way (in simple terms):
```python
@g
def f():
    ...
```
Suppose we are working TDD. Suppose that I fully tested `f` when `g` does nothing (e.g. in our case, if the repo is not remote) and I tested that when I decorate any function with `g`, everything works as expected. So, I can trust `f` and `g` taken singularly. What is the "right" way to ensure that `f` has been actually decorated with `g`? The tests I have before would pass whether `f` is decorated or not. What is the smallest test I need to add that would fail, and therefore would force me to decorate `f` with `g`?

So far the only answer to this question seemed to be `inspect.getsource` and check that there is a decorator. But I find this approach of parsing the source code a little weird...

One thing more (I hope not to confuse you, please just skip it if this is the case): I talked about TDD according to the London school. If we take the Chicago school, then probably we could say "we don't check the decorators with unittest because they are private methods, and rather check the functionalities in the public ones (`api`)." In practice this would mean to leave the tests of the `api` as they were, and delete the tests of the decorators in `git_utils`. Even though this may sound simple, it actually would lead to a higher complexity if one wants to test everything properly (one should check all possible combinations of inputs, and by adding `remote_or_not`, `commit`, and `push`, means 8 times more test). Much easier to test 2 cases (`remote_or_not`), then other 2 cases (`commit`), then other 2 cases (`push`), separately, and then make sure that they combine properly.

In short, if we change the decorators to context managers (notice: external to `api` and `registry`, so not by including them in `GitRegistry`), tests may get simpler. At the price of changing the code. If someone has a good idea to solve the testing problem I explained above with `f` and `g`, that would be ideal...
With this said, I have nothing against "moving the target" and decorate `GitRegistry` instead of the `api` functions, if this helps performance etc.

francesco086: My suggestion, considering all:
- move the remote repo functionalities into `GitRegistry` directly (so, suggestion from @skshetry will be realized)
- I am unsure whether to keep decorators or witch to context managers. When I am unsure, I prefer to be lazy, so eventually I think I would stick to the decorators.

But I leave the decision to you :)

What about grouping tests into classes, is that ok? Like in here: https://github.com/iterative/gto/pull/296/files#diff-dca8d5901c13c8fdb3912ed1fa77246f037bdda082b903fac553851476779773R31

francesco086:
...aaaaaand I think I got it! This is how you can assert that a decorator has been called: https://github.com/francesco086/gto/blob/test_decorator_call/tests/fp_test.py#L17

I don't know what do you think. In my opinion, it was interesting to see, but I am getting the strong impression that python is not really functional programming friendly :/ For each function you need its "core" implementation separated, eventually using the decorator itself would not work, because you need to be able to reference the non-decorated function....

Shall we switch to object oriented (and in this case, context managers)?

aguschin:
Sorry, I'm not that fast 😂 was trying to understand your examples 😅 

1. Thanks for thorough explanation about why tests are complex. I checked out your examples and it indeed looks like we lack some simple ways to check that functions are decorated.
2. Looks like moving clone logic to GitRegistry.from_repo should be easier to test, since we don't need to test all API methods I believe.
3. Your example with taking functional approach to its limits is very interesting indeed. And the last one with testing that the function is decorated - as well. IMHO, at this level of complexity I'd prefer this to be done with context managers. (TBH, from my side this is just the lack of knowledge on how to do it properly in functional style).
4. About joining tests in classes. Sure, you can do it like that is that's easier. I hope at some point I'll refactor all the tests to be written in the same fashion - but definitely not today 😂 

> TDD according to the London school. If we take the Chicago school

Could you maybe share something for a novice to read on that? Would be great to get some basic knowledge about TDD and these nuances, if they're important :)

francesco086:
> 2\. Looks like moving clone logic to GitRegistry.from_repo should be easier to test, since we don't need to test all API methods I believe.

Hi @aguschin , bad news about it, I believe we forgot the reason why we opted to modify the api XD
I started making the change aaaand... have a look here:
https://github.com/francesco086/gto/blob/clone_in_git_registry/gto/registry.py#L61

If we use the `with cloned_git_repo(...)` statement as part of `GitRegistry.from_repo(...)`, the tmp_dir where the cloned repo is will be deleted before anything else is run (e.g. `reg.check_ref(...)`).

We have discussed the hack of using yield instead of return [here](https://github.com/francesco086/gto/blob/clone_in_git_registry/gto/registry.py#L65), but, as you may remember, there is no guarantee that the tmp dir won't be deleted while e.g. `reg.check_ref(...)` is running... ([reference](https://stackoverflow.com/questions/41881731/is-it-safe-to-combine-with-and-yield-in-python)).

So, to make it work with the GitRegistry, we should introduce a persistent cache, which, we agreed, we currently don't want.

francesco086:
In my stubborness, I came across at again another "discovery" about decorators. I was wondering: can you reference a non-decorated function? How? Here it is:
```python
from functools import wraps

def g(f):
    @wraps(f)
    def wrapped_f(*args, **kwargs):
        print("decorated!")
        return f(*args, **kwargs)

    wrapped_f.no_decorator = f
    return wrapped_f

@g
def f():
    print("I am f")

f()
>> decorated!
>> I am f

f.no_decorator()
>> I am f
```

I believe this could be the solution to my test issues :D

francesco086:
...and when you are stubborn enough to go check the python-lang source code, you find out that the reference to the original function (the `no_decorator`) is done by  `functools.wraps` [here](https://github.com/python/cpython/blob/3.11/Lib/functools.py#L61), it's `__wrapped__`. So:
```python
from functools import wraps

def g(f):
    @wraps(f)
    def wrapped_f(*args, **kwargs):
        print("decorated!")
        return f(*args, **kwargs)

    return wrapped_f

@g
def f():
    print("I am f")

f.__wrapped__()
>> I am f
```

Here we go! I think at the end of this journey I should write an article about testing when using decorators XD

francesco086:
The other piece I needed for what I wanted to do (actually, the most important), a `get_applied_decorators` function.
Showcase:
```python
import ast
import inspect
from dataclasses import dataclass
from functools import wraps
from typing import Callable


def g(a):
    def wrap(f):
        @wraps(f)
        def wrapped_f(*args, **kwargs):
            return f(*args, **kwargs)
        return wrapped_f
    return wrap


@g("x")
@g(a="x")
def f(x):
    pass


@dataclass
class AppliedDecorator:
    f: str
    args: list
    kwargs: dict

def get_applied_decorators(f: Callable):
    o = ast.parse(inspect.getsource(f))
    ast_f = [c for c in ast.iter_child_nodes(o) if c.name == f.__name__][0]
    return [
        AppliedDecorator(f=d.func.id,
                         args=[a.value for a in d.args],
                         kwargs={k.arg: k.value.value for k in d.keywords})
        for d in ast_f.decorator_list
    ]


dec = get_applied_decorators(f)
print(dec)

>> [AppliedDecorator(f='g', args=['x'], kwargs={}), AppliedDecorator(f='g', args=[], kwargs={'a': 'x'})]
```

Unfortunately it requires parsing the source code :( But thanks to the module `ast` it is actually reasonable. Just quite some boilerplate code. I must say that I find strange that such functionalities are not built-in in python, perhaps in the inspect module...

francesco086:
Hi @aguschin , I started working out the alternative approach, here a draft PR where you can have a look at the changes:
https://github.com/francesco086/gto/pull/3

Notice how many rather intrusive changes were required. The most annoying part is `init_index_manager`, which I only started to address.

Notice also that the same exact changes were required for `GitRegistry` and `EnrichmentManager`. To avoid code duplication I introduced `FromRemoteRepoMixin`, which is basically the equivalent of `@clone_on_remote_repo`, just less general.

How does it look in your opinion? Do you think it's a promising direction?

francesco086:
I close this one in favor of https://github.com/iterative/gto/pull/309

PR #297, opened 2022-10-28
cli help mention experimental support for remote repos

francesco086:

@aguschin I remember you mentioned that you would like to document it in the README...? where/how? I could not find a good place.

### DIFF
--- a/gto/cli.py
+++ b/gto/cli.py
@@ -218,7 +218,7 @@ option_repo = Option(
     ".",
     "-r",
     "--repo",
-    help="Repository to use",
+    help="Repository to use (experimental support for remote repos)",
     show_default=True,
 )
 option_all_branches = Option(
--- a/gto/cli.py
+++ b/gto/cli.py
@@ -218,7 +218,7 @@ option_repo = Option(
     ".",
     "-r",
     "--repo",
-    help="Repository to use (experimental support for remote repos)",
+    help="Local or remote repository",
     show_default=True,
 )
 option_all_branches = Option(


gto/cli.py
```diff
@@ -218,7 +218,7 @@ def GTOGroupSection(section):
     ".",
     "-r",
     "--repo",
-    help="Repository to use",
+    help="Repository to use (experimental support for remote repos)",
```
aguschin, line 221:
```suggestion
    help="Repository to use (+ experimental support for remote repos)",
```

aguschin: btw, we say it everywhere when we have `--commit` and `--push`. Do we need to mention it here as well? IIRC, the biggest part of problems could arise is when you have a local repo and do something (i.e. non-consistency between local and remote). Not when you use the remote repo in `--repo`, cause then the problem frequency should be negligible. 
thus, should we add "experimental" word here at all? Maybe just "Repository to use (local or remote)" or something?

francesco086: Perhaps I am missing something: will the "+" do something, or the text will be rendered exactly as written: `(+ experimental support ...`?
I have never seen this convention before.

francesco086:
> Was busy with the 0.3.0 release for MLEM. Finally, it's out

Congrats!!! 🥳

gto/cli.py
aguschin: Yes, never seen this convention as well. Just thought one may assume we support remotes only, that's why I suggested `+`. But anyway, would you be ok with removing `experimental` word at all here?

francesco086: oh ok, that's what you meant :)
I think it's a bit unclear...

No problem about removing the `experimental`, but then we should remove that word also from the `commit` and `push` options, right? Shall I do it as part of this PR?

What about renaming the whole into `Local or remote repository` (`to use` perhaps is obvious and can be omitted)?

aguschin: `Local or remote repository` is fine to me. As I said above, working with remote repos look robust to me (it shouldn't cause errors 99% of the time). But working with local repos + `--push`/`--commit` looks unstable (there could be some errors - that's why I suggested adding `experimental` in first place). So I'm for keeping `experimental` there, and not using `experimental` here for `--repo` option.

aguschin: Do you agree / disagree? :)

francesco086: I understand your concern, so I agree.

Changed: https://github.com/iterative/gto/pull/297/commits/0097ea68c299ec4e85091c25729e89e368762e38

aguschin (approved):

Issue #294, opened 2022-10-24
`gto --help` sections doesn't show up


aguschin:
When installing with `pip install gto`:
```shell
(.venv) tmp %f $ gto -h                                                                                                                          10.09s

 Usage: gto [OPTIONS] COMMAND [ARGS]...


 Git Tag Ops. Turn your Git repository into an Artifact Registry:
 * Registry: Track new artifacts and their versions for releases and significant
 changes.
 * Lifecycle Management: Create actionable stages for versions marking status of
 artifact or it's readiness to be consumed by a specific environment.
 * GitOps: Signal CI/CD automation or other downstream systems to act upon these
 new versions and lifecycle updates.
 * Enrichments: Annotate and query artifact metadata with additional information.

╭─ Options ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ --version            Show version and exit                                                                                                           │
│ --verbose  -v        Print debug messages                                                                                                            │
│ --help     -h        Show this message and exit.                                                                                                     │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
╭─ Commands ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ annotate         Update artifact metadata annotations                                                                                                │
│ assign           Assign stage to specific artifact version                                                                                           │
│ check-ref        Find out the artifact version registered/assigned with ref                                                                          │
│ deprecate        Deprecate artifact, deregister a version, or unassign a stage                                                                       │
│ describe         Display enrichments for an artifact                                                                                                 │
│ history          Show a journal of registry operations                                                                                               │
│ register         Create an artifact version to signify an important, published or released iteration                                                 │
│ remove           Remove the enrichment for given artifact                                                                                            │
│ show             Show the registry state, highest version, or what's assigned in stage                                                               │
│ stages           Print list of stages used in the registry                                                                                           │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

deps
```shell
(.venv) tmp %f $ pip freeze | grep click                                                                                                         17.47s
click==8.1.3
(.venv) tmp %f $ pip freeze | grep typer                                                                                                          577ms
typer==0.6.1
```

Commands should be broken into sections like:
```shell
(gto) example-gto (main) $ gto -h                                                                                                                 146ms
Usage: gto [OPTIONS] COMMAND [ARGS]...

[...]

PR #299, opened 2022-11-08
gto --help sections doesn't show up (#294)

StepanBakshayev:
* override method GtoGroup.format_help for skipping using TyperGroup.format_help which has side-effect

### DIFF
--- a/gto/cli.py
+++ b/gto/cli.py
@@ -170,6 +170,12 @@ class GtoGroup(GtoCliMixin, TyperGroup):
                     with formatter.section(gettext(section)):
                         formatter.write_dl(sections[section])
 
+    # The typer overrides this with some trick dependent on environment,
+    # use click method for taking affect of Self.format_commands
+    def format_help(self, ctx: click.Context, formatter: click.HelpFormatter) -> None:
+        # pylint: disable=bad-super-call
+        super(TyperGroup, self).format_help(ctx, formatter)
+
     def get_command(self, ctx: Context, cmd_name: str) -> Optional[Command]:
         cmd = super().get_command(ctx, cmd_name)
         if cmd is not None:


aguschin:
@StepanBakshayev, I see some tests fail. Could you please TAL?

StepanBakshayev:
I used correctly and intentionally [super()](https://docs.python.org/3.7/library/functions.html#super). [Pylint sees](https://github.com/iterative/gto/actions/runs/3417390674/jobs/5692703957#step:7:28) this as [anti-pattern for no reason](https://docs.pylint.org/features.html#id15):

```
class GtoCommand(GtoCliMixin, TyperCommand):
...
    # The typer overrides this with some trick dependent on environment,
    # use click method for taking affect of Self.format_commands
    def format_help(self, ctx: click.Context, formatter: click.HelpFormatter) -> None:
        super(TyperGroup, self).format_help(ctx, formatter)

gto/cli.py:176:8: E1003: Bad first argument 'TyperGroup' given to super() (bad-super-call)
```

Again, simple thing connects to some strategic decisions. At first there was dependencies, now it is about linter. There are alternatives to throw away this patch or to suspend pylint diagnostic for this line. This is out of my role as one-shot contribution from stranger to give answer.

StepanBakshayev:
I guess you have more important tasks. This is a reason I did not use mentioning before. I really want to push this issue to the end.

@aguschin, what is the best way to avoid pylint claim: suspend rule for this particular place or find another code for the same effect to trick pylint?

aguschin:
Sorry @StepanBakshayev, I was sick all the previous week :(

Checking this out...

StepanBakshayev:
I see people have been spending years for merging. Week is nothing. I got. I will add suspending directive for pylint.

mike0sv:
pylint probably wants you to use super like this `super().format_help(...)`

aguschin (approved):

Issue #377, opened 2023-07-25
Bug: gto irresponsive


francesco086:
## Background
I have a dvc/gto repository (unfortunately private, cannot share it). Here the most recent tags:

![Screenshot 2023-07-25 at 15 02 34](https://github.com/iterative/gto/assets/8321888/5fecec56-f32d-4b35-98f0-46b7176582da)

I tried to register a new version and I got an error. Then I tried to do a `gto show`, and that also result in the same problem...

![Screenshot 2023-07-25 at 15 07 40](https://github.com/iterative/gto/assets/8321888/269ebd76-1de1-44fe-94ba-2dec4e00e652)

![Screenshot 2023-07-25 at 15 08 07](https://github.com/iterative/gto/assets/8321888/d9d7f61a-5102-4565-a142-fa736305a2d6)

I tried with gto 0.3.5, 0.3.3, 0.3.1.

Here the python (3.11.1) packages:
```
Package                                 Version
--------------------------------------- -----------
adal                                    1.2.7
adlfs                                   2023.4.0
aiohttp                                 3.8.4
aiohttp-retry                           2.8.3
aiosignal                               1.3.1
amqp                                    5.1.1
antlr4-python3-runtime                  4.9.3
appdirs                                 1.4.4
applicationinsights                     0.11.10
argcomplete                             2.1.2
async-timeout                           4.0.2
asyncssh                                2.13.1
atpublic                                3.1.1
attrs                                   23.1.0
azure-appconfiguration                  1.1.1
azure-batch                             13.0.0
azure-cli                               2.48.1
azure-cli-core                          2.48.1
azure-cli-telemetry                     1.0.8
azure-common                            1.1.28
azure-core                              1.26.4
azure-cosmos                            3.2.0
azure-data-tables                       12.4.0
azure-datalake-store                    0.0.53
azure-graphrbac                         0.60.0
azure-identity                          1.13.0
azure-keyvault                          1.1.0
azure-keyvault-administration           4.0.0b3
azure-keyvault-keys                     4.8.0b2
azure-loganalytics                      0.1.1
azure-mgmt-advisor                      9.0.0
azure-mgmt-apimanagement                3.0.0
azure-mgmt-appconfiguration             3.0.0
azure-mgmt-appcontainers                2.0.0
azure-mgmt-applicationinsights          1.0.0
azure-mgmt-authorization                3.0.0
azure-mgmt-batch                        17.0.0
azure-mgmt-batchai                      7.0.0b1
[...]

PR #379, opened 2023-07-31
Expand the ManyVersions error message

francesco086:
Contributes to https://github.com/iterative/gto/issues/377

### DIFF
--- a/gto/exceptions.py
+++ b/gto/exceptions.py
@@ -83,7 +83,7 @@ class VersionRequired(GTOException):
 
 
 class ManyVersions(GTOException):
-    _message = "'{versions}' versions of artifact '{name}' found"
+    _message = "Multiple versions ('{versions}') of artifact '{name}' found on the same git commit. Multiple versions are not allowed."
 
     def __init__(self, name, versions) -> None:
         self.message = self._message.format(name=name, versions=versions)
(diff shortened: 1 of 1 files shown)


gto/exceptions.py
```diff
@@ -83,7 +83,7 @@ def __init__(self, name) -> None:
 
 
 class ManyVersions(GTOException):
-    _message = "'{versions}' versions of artifact '{name}' found"
+    _message = "Multiple versions ('{versions}') of artifact '{name}' found on the same git commit. Multiple versions are not allowed."
```
shcheklein, line 86: nit: Multiple versions are not allowed on the same Git commit. Found <versions> of an artifact <name> on <git ref>

wdyt?

shcheklein (approved):
A minor comment. Thanks @francesco086 . Overall we can go as is. Let me know.

gto/exceptions.py
francesco086: Sorry, I am not able to understand your suggestion: "Found of an artifact on"?

Would this be ok?
`Multiple versions are not allowed on the same Git commit. Violation detected for artifact '{name}', versions '{versions}'.`

francesco086: Just in case it is good, I pushed the change.

shcheklein: thanks!

~/Desktop/foo
pmrowla, line 1: @francesco086 was this addition intentional?

francesco086: Of course. Not. 😅
Sry about that

PR #398, opened 2023-08-29
change artifact name regex

dberenbaum:
~Having an issue with tests, but opening this for discussion in the meantime.~

The PR adds support for uppercase letters and `_`. I decided against support for `.` because there are some restrictions around how it can be used in [git refs](https://git-scm.com/docs/git-check-ref-format). 

Related: https://github.com/iterative/dvc/issues/9821

TODO:
- [x] Tests
- [ ] DVC PR
- [ ] DVCLive PR to improve [warning](https://github.com/iterative/dvclive/blob/95000f45638d5126b7d76833050f549167a7449d/src/dvclive/live.py#L462-L466)
- [ ] Docs PR

### DIFF
--- a/gto/constants.py
+++ b/gto/constants.py
@@ -42,7 +42,7 @@ def tag_to_name(value):
 
 
 dirname = "[a-z0-9-_./]+"  # improve?
-name = r"[A-Za-z0-9]([A-Za-z0-9-/_]*[A-Za-z0-9])?"
+name = r"[a-z0-9]([a-z0-9-/_]*[A-Za-z0-9])?"
 semver = r"(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?"
 counter = "?P<counter>[0-9]+"
 name_re = re.compile(f"^{name}$")
--- a/gto/constants.py
+++ b/gto/constants.py
@@ -42,7 +42,7 @@ def tag_to_name(value):
 
 
 dirname = "[a-z0-9-_./]+"  # improve?
-name = r"[a-z0-9]([a-z0-9-/]*[a-z0-9])?"
+name = r"[A-Za-z0-9]([A-Za-z0-9-/_]*[A-Za-z0-9])?"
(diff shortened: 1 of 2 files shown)


mnrozhkov (commented):
Thanks @dberenbaum! Looks good!

dberenbaum:
@aguschin Do you mind reviewing when you have a chance?

gto/constants.py
```diff
@@ -42,7 +42,7 @@ def tag_to_name(value):
 
 
 dirname = "[a-z0-9-_./]+"  # improve?
-name = r"[a-z0-9]([a-z0-9-/]*[a-z0-9])?"
+name = r"[A-Za-z0-9]([A-Za-z0-9-/_]*[A-Za-z0-9])?"
```
shcheklein, line 45: not sure about modern Git, but In general the recommendation is to avoid using upper case tag / branch names (since Windows and Mac could mess up with that). Let's just add some symbols for now? Did we see people complaining? We could also automatically convert for simplicity / usability when people manipulate / pass those names in UI / CLI.

(same btw applies to the semver regex that I see is also using `A-Z`

wdyt?

dberenbaum: No strong opinion from me (Git does handle casing from what research I did but it's obviously not the typical convention). @mnrozhkov You had asked about this, right? Do you find it useful/common to need uppercase letters?

aguschin: +1 to Ivan's comment here. I think the simpler it is, the less changes to make a mistake (like mistake `i`.upper() from `L`.lower()), etc. That was original intention.

mnrozhkov: > No strong opinion from me (Git does handle casing from what research I did but it's obviously not the typical convention). @mnrozhkov You had asked about this, right? Do you find it useful/common to need uppercase letters?

I think we should support uppercase letters for artifact names! 

[Here](https://iterativeai.slack.com/archives/CB41NAL8H/p1691411376466919) I raised a discussion because of an error by calling `live.log_artifact("scalarTransportFoam")` 
- The "scalarTransportFoam" is a directory name generated by third-party software (simulation). There is no option to name it by the DVC user

dberenbaum: @mnrozhkov Do you think it's important to keep the casing, or do you think the suggestion to force the name to lowercase would be acceptable?

mnrozhkov: @dberenbaum as I answered above: 
> I think we should support uppercase letters for artifact names!

It's not acceptable for the customer I mentioned to change the name to lowercase 
1. Because this name is generated by a third-party software, it's standard 
2. This name may be used by the same software or legacy code, so it could break the workflow 

I feel it's fine to keep lowercase letters for git tag/branch things, but should we be so strict for artifact names? 
Feel free to have a call if I miss smth here.

dberenbaum: Do others feel it's a blocker? I don't see a strong reason not to allow it even if it's not the typical recommended pattern.

shcheklein: > I feel it's fine to keep lowercase letters for git tag/branch things, but should we be so strict for artifact names?
Feel free to have a call if I miss smth here.

So, is it fine to "complain" then if someone is trying to register an artifact with a different name?

I'm fine to keep the change, also seems reasonable to allow this on the dvc.yaml level. I would still probably convert this into lowercase when we do tags. Just to be safe here.

dberenbaum: It's going to expand the scope here to have special logic for how to handle uppercase letters, so I'm going to drop that from the PR and let's keep a separate issue to discuss it.

shcheklein (approved):

Issue #400, opened 2023-09-05
update GTO to be in line with iterative/py-template standards


pmrowla:
DVC officially, supports Python 3.8 through 3.11 but GTO CI is only run against 3.8 through 3.10 (GTO CI fails if you enable 3.11). As far as I can tell, the 3.11 failure is because GTO requires outdated pylint and mypy installations that are just broken on 3.11. The actual test suite passes in 3.11.

IMO there is a lot more work that needs to be done in GTO to bring it in line with current DVC and `iterative/py-template` standards, but for the time being I am leaving all of that alone (GTO codebase fails pylint/mypy checks after updating and enabling the py-template or DVC configurations for those tools)

Making GTO pass current mypy/pylint may also require updating other deps (and updating pydantic may require migrating from pydantic v1 to v2 - https://github.com/iterative/gto/pull/375)

_Originally posted by @pmrowla in https://github.com/iterative/gto/issues/378#issuecomment-1672866569_

shcheklein:
I think we've updated pydantic, also it runs on 3.11. What else do we usually need to get it to py-template?

pmrowla:
This can be closed, issues were all addressed in:
- https://github.com/iterative/gto/pull/408
- https://github.com/iterative/gto/pull/410
- https://github.com/iterative/gto/pull/413

shcheklein:
Thanks, Peter. Do we want / need to upgrade it to py template?

pmrowla:
There isn't really a straightforward way to retroactively apply a cookiecutter/cruft template on top of an existing git repo, so any major changes that are made in py-template in the future would need to be done manually in GTO.

PR #401, opened 2023-09-05
Bump pylint, pydantic, remove "artifacts.yaml" logic

aguschin:
👀 

This will require Studio BE update - we need to remove old logic that works with `artifacts.yaml` - that will make Studio MR backend codebase more easy to work with! To clarify, before we already removed "write" access to old `artifacts.yaml`, but kept "read" access. Now we remove "read" access as well. (Please notice, that will lead to some users losing info about model versions that are 3+ months old, but I hope that's ok).

So once this merged, consider bumping major or minor version, but not the patch.


aguschin:
cc @amritghimire about changes required in BE ^

aguschin:
@pmrowla, I assume this is what you meant to make it compatible with py-template. There are few things still that should be fixed - like remove top-level silence statements here https://github.com/iterative/gto/blob/bumping-versions/gto/config.py#L1, but overall it should be ready for py-template I hope!

aguschin:
close https://github.com/iterative/gto/issues/400

shcheklein:
Since studio depends for now on MLEM and MLEM depends on the first Pydantic, will those co-exist fine together or does it means we need to upgrade / drop MLEM dependency?

.github/workflows/check-test-release.yml
```diff
@@ -37,7 +37,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python: ["3.8", "3.9", "3.10"]
+        python: ["3.9", "3.10", "3.11"]
```
skshetry, line 40: DVC still supports 3.8 so I don't think we should drop that here too.

skshetry:
We discussed this in today's team meeting and decided not to merge this PR due to the breakages involved. We don't want to risk merging this for now. 

Most of the pylint/mypy changes has been cherry-picked into the main branch, and CI is now running from Python 3.8 - 3.11 versions now. 

I have created #413 that supports both Pydantic v1 and v2, so there's no immediate need to migrate to use pydantic v2 APIs too.

PR #408, opened 2023-09-14
run on py311

skshetry:


pmrowla:
gto CI linter steps fail in 3.11 (see https://github.com/iterative/gto/issues/400)

skshetry:
@pmrowla , I have updated from py-template.

gto/cli.py
```diff
@@ -657,7 +657,9 @@ def check_ref(
     ), "Only one output formatting flags is allowed"
     result = gto.api.check_ref(repo, ref)
     if len(result) > 1:
-        NotImplementedInGTO("Checking refs that created 1+ events is not supported")
+        raise NotImplementedInGTO(
```
skshetry, line 660: Looks like this is a bug, where an exception was not being raised.

PR #413, opened 2023-09-14
support pydantic v2

skshetry:
Also supports pydantic v1.

pydantic v2 provides v1 api through `pydantic.v1` import, so we can support both versions by trying to import from `pydantic.v1` and fallback to old behaviour.

See https://docs.pydantic.dev/latest/migration/#continue-using-pydantic-v1-features.

pydantic == 2.0.0 does not work due to https://github.com/pydantic/pydantic/issues/6361, so we need to blacklist that specific version.

### DIFF
--- a/gto/config.py
+++ b/gto/config.py
@@ -3,13 +3,13 @@ import pathlib
 from pathlib import Path
 from typing import Any, Dict, List, Optional
 
-from pydantic import BaseModel, BaseSettings, validator
-from pydantic.env_settings import InitSettingsSource
 from ruamel.yaml import YAML
 
 from gto.constants import assert_name_is_valid
 from gto.exceptions import UnknownStage, UnknownType, WrongConfig
 from gto.ext import EnrichmentReader, find_enrichment_types, find_enrichments
+
+from ._pydantic import BaseModel, BaseSettings, InitSettingsSource, validator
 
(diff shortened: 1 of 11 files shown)


gto/config.py
```diff
@@ -41,11 +41,13 @@ class Config:
 
     def assert_type(self, name):
         assert_name_is_valid(name)
+        # pylint: disable-next=unsupported-membership-test
```
skshetry, line 44: This and the following is needed due to a false-positive with pylint.

shcheklein:
@skshetry is there a reason to support both?

skshetry:
> @skshetry is there a reason to support both?

Studio still relies on MLEM and therefore on Pydantic v1. Also I did not want to block any DVC users that are using pydantic v2 (or, pydantic v1 even). 

Also, using pydantic v2 APIs means changing code, whereas if we use v1 APIs for now, we can support both and is working perfectly fine.

shcheklein:
sounds good. can we run tests with both then?

skshetry:
I consider pydantic v1 support as best effort from now on. I don’t think we need to test both.

shcheklein (approved):

shcheklein:
sound good, I think Studio has enough tests to detect failure if there is a regression in pydantic v1

pmrowla (approved):

PR #416, opened 2023-09-15
Revert "chore(deps): update codecov/codecov-action action to v4"

skshetry:
Reverts iterative/gto#414


skshetry:
Looks like they accidentally tagged v4. See https://github.com/codecov/codecov-action/issues/1089

PR #417, opened 2023-09-19
fix(registry): push to remote automatically only on cloned repos

### LINKED ISSUE #405: gto should not push on register without --push flag.
The gto command has a --push flag which according to the [docs](https://mlem.ai/doc/gto/command-reference/register) will push a created tag.

`--push - Push created tag automatically (experimental)` 

However, looking into the [code ](https://github.com/iterative/gto/blob/3b3dec14ca0f5ab4c2fdcbad8740322d4ce189b7/gto/api.py#L100C28-L100C29)the tag will automatically be pushed if the git repository has a remote.

This is unexpected as it is not a mentioned actions in the docs. I looked back into a older versions of gto (v0.3.5) and saw that there was similar logic, but found the logic to not be implemented/functional. 

I have a github action that automatically increments the tag version when changes are pushed to a branch (prod). 

```
name: Increment Tag Version
on:
  push:
    branches:
      - prod

jobs:
  tag-up:
    environment: ${{ github.ref_name }}
    runs-on: ubuntu-latest

    steps:
      - uses: actions/checkout@v3
        with:
          token: ${{ secrets.<Personal Access Token>}}
          fetch-depth: 0

      - name: ⬆️ tag version ⬆️
        run: |
          pip install gto
          git config --local user.email "41898282+github-actions[bot]@users.noreply.github.com"
          git config --local user.name "github-actions[bot]"
          gto register pfc-irm HEAD --bump-patch
    
      - name: ✍ push tags ✍
        run: |
          git push origin --tags
```

This action was working with gto prior to the version 1 release. It now fails with the logs:

```
Created git tag '<model>@v1.0.11' that registers version
Running `git push origin <model>@v1.0.11`
❌ The command `git push origin <model>@v1.0.11` failed. Make sure your local 
repository is in sync with the remote.
```

I am not sure why the automatic gto register push is failing; however, I would suggest that the push only happens when the `--push` flag is provided. An additional `--auto-push` flag could be created that could cover the current behavior. 

Alternatively pushing could be the default behavior and a `--no-push` flag could be provided to prevent pushing.

shcheklein:
Fixes #405 

We've changed the semantic of the operation during the recent migration to `scmrepo`. The key thing is the `has_remote(reg.scm)` calls and implementation. I think the original intention was to push automatically when we run GTO operations on a remote repo (means we are cloning it into a temp dir, do some op, and push the result back).

## TODO

- [x] Tests
- [x] Review CLI option description
- [x] Review docs

## Docs

Relevant docs update is here https://github.com/iterative/dvc.org/pull/4879

### DIFF
--- a/gto/api.py
+++ b/gto/api.py
@@ -19,7 +19,6 @@ from gto.constants import (
     parse_shortcut,
 )
 from gto.exceptions import NoRepo, NotImplementedInGTO, RefNotFound, WrongArgs
-from gto.git_utils import has_remote
 from gto.index import Artifact, RepoIndexManager
 from gto.registry import GitRegistry
 from gto.tag import parse_name as parse_tag_name
@@ -97,7 +96,7 @@ def register(
             bump_major=bump_major,
             bump_minor=bump_minor,
             bump_patch=bump_patch,
-            push=push or has_remote(reg.scm),
+            push=push,
             stdout=stdout,
             author=author,
             author_email=author_email,
@@ -131,7 +130,7 @@ def assign(
             message=message,
             simple=simple,
             force=force,
-            push=push or has_remote(reg.scm),
+            push=push,
             skip_registration=skip_registration,
             stdout=stdout,
             author=author,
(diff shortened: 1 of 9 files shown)


pmrowla:
LGTM, I'm ok with merging this and doing tests in a follow up if a customer needs the fix

shcheklein:
For the record - added tests, updated docs before merging this. It was approved by Peter.

PR #425, opened 2023-10-09
fix(tags): parse timezone information from a tag

shcheklein:
We don't parse timezones, means we are not getting right dates in Studio. Also Studio tests produce a lot of warning about "naive" date times.

Not critical, but better to fix in advance.

### DIFF
--- a/gto/tag.py
+++ b/gto/tag.py
@@ -1,6 +1,6 @@
+import datetime
 import os
 import re
-from datetime import datetime
 from enum import Enum
 from typing import FrozenSet, Iterable, Optional, Union
 
@@ -124,7 +124,7 @@ class Tag(BaseModel):
     name: str
     version: Optional[str]
     stage: Optional[str]
-    created_at: datetime
+    created_at: datetime.datetime
     tag: GitTag
 
     class Config:
@@ -134,7 +134,7 @@ class Tag(BaseModel):
 def parse_tag(tag: GitTag):
     return Tag(
         tag=tag,
-        created_at=datetime.fromtimestamp(tag.tag_time),
+        created_at=tag.tag_datetime,
         **parse_name(tag.name),
     )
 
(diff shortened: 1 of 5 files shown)


gto/tag.py
```diff
@@ -124,17 +124,18 @@ class Tag(BaseModel):
...
-    created_at: datetime
+    created_at: datetime.datetime
     tag: GitTag
 
     class Config:
         arbitrary_types_allowed = True
 
 
 def parse_tag(tag: GitTag):
+    tz = datetime.timezone(datetime.timedelta(minutes=int(tag.tag_time_offset)))
```
shcheklein, line 135: C: Offset come in a form of `540`, `-180`, etc. @pmrowla can we confirm that it is always the case? and that we are always getting `tag_time_offset` in the GitTag?

pmrowla: The timezone offset should be in seconds, looks like one of the scmrepo backends is returning minutes (which is an scmrepo bug)

gto/tag.py
```diff
@@ -124,17 +124,18 @@ class Tag(BaseModel):
...
     return Tag(
         tag=tag,
-        created_at=datetime.fromtimestamp(tag.tag_time),
+        created_at=datetime.datetime.fromtimestamp(tag.tag_time, tz=tz),
```
pmrowla, line 138:
```suggestion
        created_at=tag.tag_datetime,
```

gto/tag.py
pmrowla: scmrepo 1.4.0 is released with the fix for this, and you can also now use `tag.tag_datetime` to get the timezone-aware python datetime

shcheklein: thanks @pmrowla !

pmrowla:
We should also bump the scmrepo minver to 1.4.0 in setup.py, other than that LGTM

PR #426, opened 2023-10-10
Minor correction to error message

tapadipti:
Related to https://github.com/iterative/dvc/issues/9821

### DIFF
--- a/gto/constants.py
+++ b/gto/constants.py
@@ -71,7 +71,7 @@ def check_string_is_valid(value, regex=name_re):
 def assert_name_is_valid(value):
     if not check_string_is_valid(value, regex=name_re):
         raise ValidationError(
-            f"Invalid value '{value}'. Only lowercase english letters, , '-', '/' are allowed."
+            f"Invalid value '{value}'. Only english letters, '_', '-', '/' are allowed."
             "Value must be of len >= 2, must with a letter and end with a letter or a number."
         )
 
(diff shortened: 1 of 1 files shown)


gto/constants.py
```diff
@@ -71,7 +71,7 @@ def check_string_is_valid(value, regex=name_re):
 def assert_name_is_valid(value):
     if not check_string_is_valid(value, regex=name_re):
         raise ValidationError(
-            f"Invalid value '{value}'. Only lowercase english letters, , '-', '/' are allowed."
+            f"Invalid value '{value}'. Only english letters, '_', '-', '/' are allowed."
             "Value must be of len >= 2, must with a letter and end with a letter or a number."
```
dberenbaum, line 75:
```suggestion
            f"Invalid value '{value}'. Only letters, numbers, '_', '-', '/' are allowed."
            "Value must be of len >= 2 and must start and end with a letter or a number."
```

gto/constants.py
```diff
@@ -80,7 +80,7 @@ def assert_fullname_is_valid(value):
     if not check_string_is_valid(value, regex=fullname_re):
         # fix error message to be regex-specific
         raise ValidationError(
-            f"Invalid value '{value}'. Only lowercase english letters, , '-', '/' are allowed."
+            f"Invalid value '{value}'. Only english letters, '_', '-', '/' are allowed."
             "Value must be of len >= 2, must start with a letter and end with a letter or a number."
```
dberenbaum, line 84:
```suggestion
            f"Invalid value '{value}'. Only letters, numbers, '_', '-', '/' are allowed."
            "Value must be of len >= 2 and must start and end with a letter or a number."
```

dberenbaum (approved):

PR #444, opened 2024-02-05
chore(deps): drop 3.8 support

shcheklein:
Drop since `scmrepo` dropped it, `dvc`, etc.

Needs to fix CI and start fixing the update of the `scmrepo` here https://github.com/iterative/gto/pull/443


setup.py
```diff
@@ -18,7 +18,7 @@
 
 
 tests = [
-    "pytest",
+    "pytest<8",
```
shcheklein, line 21: waiting for this it seems https://github.com/TvoroG/pytest-lazy-fixture/issues/65

PR #447, opened 2024-03-25
chore: get rid of setup.py/setup.cfg

skshetry:
And misc maintenace updates:

* Move all of the configs to pyproject.toml instead of 3 config.
* Remove isort/black/flake8 in favor of ruff
* Use local mypy hook instead of mirrors-mypy, as it hides a lot of errors.
* Changes line-length to 88 (default in ruff). Luckily, this did not change any diff.
* Add support for Python 3.12, and run tests in CI for 3.12.
* Introduce `dev` dependency extra.

The `ruff` rules are equivalent to flake8. We can expand them in a separate PR if needed.
Also, I haven't removed pylint. If we expand `ruff` rules, we can get rid of `pylint`.
But that can be done in the next PR as my goal with this PR is to get rid of `setup.py`/`setup.cfg`.


shcheklein (approved):
thanks @skshetry , great update!

PR #478, opened 2025-08-19
chore(linters): fix mypy, ruff, pylint

shcheklein:
Housekeeping, maintenance


skshetry:
Why was 3.9 support dropped?

PR #490, opened 2026-03-18
fix: gto --plain doesnt truncate wide versions

### LINKED ISSUE #489: gto show --plain truncates latest version values in global output table (breaks CLI parsing for multi-digit semver)
### Summary

`gto show --plain` appears to truncate `latest` version values in the global summary table when the value is wide (e.g. `v0.100.0` becomes `v0.100.`).

This causes downstream automation to pick an invalid version string if it parses the global `--plain` output.

### Environment

- GTO CLI installed via `pip install gto`
- Repo using GTO tags like `ml_agent@v0.100.0`
- Observed on Linux in CI (GitHub Actions) and locally

### What I observed

Given a real tag and registration:

- Git tag exists: `ml_agent@v0.100.0`
- Service-specific output is correct:
  - `gto show ml_agent --plain` returns `v0.100.0`
- Global output is truncated:
  - `gto show --plain` shows `ml_agent  v0.100.` in `latest`

Example (global):

```text
name         latest    #production    #staging
ml_agent     v0.100.   v0.98.1        v0.99.0
api          v0.63.0   v0.62.0        v0.62.0
```

Example (service-specific):

```text
artifact    version    stage       created_at                 ref
ml_agent    v0.100.0               2026-03-18 22:33:04+00:00  ml_agent@v0.100.0
...
```

### Why this is a bug

`--plain` output is commonly used for machine parsing in CI scripts.
If the table truncates values, parsing `$2` yields an invalid version (`v0.100.`), which then fails deployment/image resolution.

### Expected behavior

One of the following:

1. `gto show --plain` should never truncate field values (machine-safe output), or
2. Provide a guaranteed machine-readable mode (e.g. JSON/CSV) for global and service-specific queries.

### Actual behavior

Global `gto show --plain` table truncates `latest` for wider values, while service-specific output does not.

### Minimal reproduction

1. Have artifact versions including at least one with a wider semver like `v0.100.0`.
2. Run:
   - `gto show --plain`
   - `gto show <artifact> --plain`
3. Compare `latest` in global output vs version in service-specific output.

### Workaround

Use service-specific query only:

- `gto show <service> --plain` and parse rows there,
- avoid parsing the global `gto show --plain` summary table.

ZivotJeKrasny:
Fixes #489

### DIFF
--- a/gto/api.py
+++ b/gto/api.py
@@ -319,7 +319,7 @@ def _show_registry(
     """Show current registry state"""
 
     def format_hexsha(hexsha):
-        return hexsha[:7] if truncate_hexsha else hexsha
+        return hexsha[:7] if truncate_hexsha and is_hexsha(hexsha) else hexsha
 
     with GitRegistry.from_url(repo) as reg:
         stages = list(reg.get_stages())
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -595,6 +595,41 @@ def test_if_unassign_with_remote_repo_then_invoke_git_push_tag(tmp_dir: TmpDir):
             )
 
 
+@pytest.mark.usefixtures("artifact")
+def test_show_registry_table_does_not_truncate_wide_semver(tmp_dir: TmpDir, scm: Git):
+    """Regression test: _show_registry must not truncate semver strings like v0.100.0.
+
+    Versions such as v0.100.0 are exactly 8 chars, exceeding the 7-char hexsha
+    truncation limit. Previously, format_hexsha in _show_registry lacked the
+    is_hexsha() guard present in _show_versions, causing the version to be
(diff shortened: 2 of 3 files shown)


gto/api.py
```diff
@@ -319,7 +319,7 @@ def _show_registry(
     """Show current registry state"""
 
     def format_hexsha(hexsha):
```
shcheklein, line 321: how / where is called outside? 

It seems the problem here is that we pass something that is _not_ hash into a function that is named `format_hexsha` . T

This guard should be outside of this function.

shcheklein (changes requested):
Thanks for your PR. Please review comment's address each meaningfully. Probably a bit broader refactoring is needed here.

shcheklein:
Closing as stale

PR #497, opened 2026-07-08
chore(deps): drop Python 3.9, add 3.14, consolidate dependency updates

shcheklein:
Consolidates all open Renovate deps/chore PRs into one change, fixes the CI breakage on `main`, and moves the supported Python range to 3.10–3.14.

## Why main was red

The weekly runs on `main` have been failing on Python 3.10+ since early June. Root cause: **typer ≥0.26 vendors click** and no longer depends on it, so fresh installs stopped pulling click in — while gto imports `click` directly (`gto/cli.py`, `gto/utils.py`, `tests/conftest.py`) without declaring it. pylint failed with `E0401: Unable to import 'click'`.

It's worse than lint: gto's CLI machinery subclasses real-click classes (`GtoCliMixin(click.Command)`, `@click.pass_context`) and mixes them into typer's (now vendored-click) `TyperCommand`/`TyperGroup` MROs. With typer 0.26.x even `gto --help` crashes on startup (`TypeError: Command.__init__() got an unexpected keyword argument 'rich_markup_mode'`) — i.e. a fresh `pip install gto` on Python 3.10+ currently yields a broken CLI. Python 3.9 was unaffected only because it resolves typer 0.23.2 (0.24+ requires Python ≥3.10; all pre-0.26 versions depend on real click).

**Fix:** declare `click>=8,<9` explicitly and pin `typer>=0.4.1,<0.26`. Migrating to vendored-click typer (≥0.26) requires reworking the CLI help/exception machinery against `typer._click` and should be tracked as a separate effort.

## Python support

- Drop Python 3.9 (EOL October 2025): removed from test matrix, classifiers; `requires-python = ">=3.10"`
- Add Python 3.14 to the test matrix and classifiers
- Remove the `CliRunner` `mix_stderr` shim in `tests/conftest.py` (only needed for the old click that py3.9 resolved)
- `zip(..., strict=False)` fixes auto-applied by ruff `B905`, which activates with the 3.10+ target

## Consolidated Renovate updates

| PR | Update | Resolution |
|---|---|---|
| #496 | `actions/checkout` v5 → v7 | taken as-is |
| #493 | `codecov/codecov-action` v4 → v7 | taken as-is |
| #484 | deploy job Python 3.13 → 3.14 | taken, plus 3.14 in test matrix |
| #487 | pylint 3.3.8 → 4.0.6 | taken (pylint 4 requires ≥3.10 — unblocked by the 3.9 drop) |
| #482 | pylint → 3.3.9 | superseded by #487 |
| #495 | mypy → v2 | delivered as `mypy==2.2.0` (the py3.9-only `<1.11.0` pin is gone with 3.9) |
| #494 | mypy → `<1.20.3` | superseded by #495 |

## Verification

Ran locally on macOS for Python **3.10**, **3.13**, and **3.14**, each with a fresh venv:

- `pip install -e .[dev]` resolves cleanly (click 8.4.2, typer 0.23.2, mypy 2.2.0, pylint 4.0.6, pygit2 1.19.3)
- `pre-commit run --all-files` — all hooks pass (ruff, ruff-format, mypy 2.2.0, pylint 4.0.6)
- `pytest` — 165 passed on each
- `gto --help` / `gto show` / `gto --version` work (broken before on 3.10+ with unpinned typer)
- `python -m build && twine check --strict dist/*` passes on 3.14 (deploy job path)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/gto/api.py
+++ b/gto/api.py
@@ -364,6 +364,7 @@ def _show_registry(
                     d["version"],
                 ]
                 + [d["stage"][name] for name in stages],
+                strict=False,
             ),
         )
         for name, d in sorted(models_state.items())
--- a/gto/api.py
+++ b/gto/api.py
@@ -364,7 +364,7 @@ def _show_registry(
                     d["version"],
                 ]
                 + [d["stage"][name] for name in stages],
-                strict=False,
+                strict=True,
             ),
         )
         for name, d in sorted(models_state.items())
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,10 +32,10 @@ dependencies = [
     "scmrepo>=3,<4",
     "semver>=2.13.0",
     "tabulate>=0.8.10",
-    # typer>=0.24 vendors click and is incompatible with gto's click-based
+    # typer>=0.26 vendors click and is incompatible with gto's click-based
     # CLI machinery (subclassing click.Command, @click.pass_context, etc.):
     # the CLI crashes on startup. Migration tracked separately.
-    "typer>=0.4.1,<0.24",
+    "typer>=0.4.1,<0.26",
 ]
 
 [project.urls]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,15 +12,17 @@ keywords = ["git", "repo", "repository", "artifact", "registry", "developer-tool
 classifiers = [
     "Development Status :: 2 - Pre-Alpha",
     "Programming Language :: Python :: 3",
-    "Programming Language :: Python :: 3.9",
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12",
     "Programming Language :: Python :: 3.13",
+    "Programming Language :: Python :: 3.14",
 ]
-requires-python = ">=3.9"
+requires-python = ">=3.10"
 dynamic = ["version"]
 dependencies = [
+    # gto imports click directly, so declare it explicitly
+    "click>=8,<9",
     "entrypoints",
     "funcy",
     "pydantic>=2",
@@ -30,7 +32,10 @@ dependencies = [
     "scmrepo>=3,<4",
     "semver>=2.13.0",
     "tabulate>=0.8.10",
-    "typer>=0.4.1",
+    # typer>=0.24 vendors click and is incompatible with gto's click-based
+    # CLI machinery (subclassing click.Command, @click.pass_context, etc.):
+    # the CLI crashes on startup. Migration tracked separately.
+    "typer>=0.4.1,<0.24",
 ]
 
 [project.urls]
@@ -48,11 +53,8 @@ tests = [
 ]
 dev = [
     "gto[tests]",
-    "mypy==1.17.1; python_version > '3.9'",
-    # mypy>=1.11.0 crashes when used with pydantic-settings on Python 3.9,
-    # see: https://github.com/python/mypy/issues/17535
-    "mypy<1.11.0; python_version <= '3.9'",
-    "pylint==3.3.8",
+    "mypy==2.2.0",
+    "pylint==4.0.6",
     "types-PyYAML",
     "types-filelock",
     "types-freezegun",
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -298,7 +298,7 @@ def test_check_ref_catch_the_bug(scm: Git):
     for assignment, tag in zip(
         [assignment1, assignment2, assignment3],
         [f"{NAME}#staging#1", f"{NAME}#prod#2", f"{NAME}#dev#3"],
-        strict=False,
+        strict=True,
     ):
         events = gto.api.check_ref(scm, tag)
         assert len(events) == 1, events
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -298,6 +298,7 @@ def test_check_ref_catch_the_bug(scm: Git):
     for assignment, tag in zip(
         [assignment1, assignment2, assignment3],
         [f"{NAME}#staging#1", f"{NAME}#prod#2", f"{NAME}#dev#3"],
+        strict=False,
     ):
         events = gto.api.check_ref(scm, tag)
         assert len(events) == 1, events
--- a/tests/test_registry.py
+++ b/tests/test_registry.py
@@ -431,6 +431,7 @@ def test_registry_state_tag_tag(tmp_dir: TmpDir):
         for appeared, expected in zip(
             iter_over(appeared_state["artifacts"][artifact]["versions"]),
             iter_over(expected_state["artifacts"][artifact]["versions"]),
+            strict=False,
         ):
             check_obj(appeared, expected, exclude["versions"])
 
@@ -438,6 +439,7 @@ def test_registry_state_tag_tag(tmp_dir: TmpDir):
                 for a, e in zip(
                     iter_over(appeared[key]),
                     iter_over(expected[key]),
+                    strict=False,
                 ):
                     check_obj(a, e, exclude[key])
 
--- a/tests/test_registry.py
+++ b/tests/test_registry.py
@@ -431,7 +431,7 @@ def test_registry_state_tag_tag(tmp_dir: TmpDir):
         for appeared, expected in zip(
             iter_over(appeared_state["artifacts"][artifact]["versions"]),
             iter_over(expected_state["artifacts"][artifact]["versions"]),
-            strict=False,
+            strict=True,
         ):
             check_obj(appeared, expected, exclude["versions"])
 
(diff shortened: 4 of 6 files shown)


gto/api.py
```diff
@@ -364,6 +364,7 @@ def format_hexsha(hexsha):
                     d["version"],
                 ]
                 + [d["stage"][name] for name in stages],
+                strict=False,
```
shcheklein: explain this change

tests/test_api.py
```diff
@@ -298,6 +298,7 @@ def test_check_ref_catch_the_bug(scm: Git):
     for assignment, tag in zip(
         [assignment1, assignment2, assignment3],
         [f"{NAME}#staging#1", f"{NAME}#prod#2", f"{NAME}#dev#3"],
+        strict=False,
```
shcheklein: explain this change

pyproject.toml
```diff
@@ -12,15 +12,17 @@ keywords = ["git", "repo", "repository", "artifact", "registry", "developer-tool
 classifiers = [
     "Development Status :: 2 - Pre-Alpha",
     "Programming Language :: Python :: 3",
-    "Programming Language :: Python :: 3.9",
```
shcheklein, line 15: @skshetry are we good to drop Python 3.9 now? should not be impacting DVC I think, right?

pyproject.toml
```diff
@@ -30,7 +32,10 @@ dependencies = [
     "scmrepo>=3,<4",
     "semver>=2.13.0",
     "tabulate>=0.8.10",
-    "typer>=0.4.1",
+    # typer>=0.24 vendors click and is incompatible with gto's click-based
+    # CLI machinery (subclassing click.Command, @click.pass_context, etc.):
+    # the CLI crashes on startup. Migration tracked separately.
+    "typer>=0.4.1,<0.24",
```
shcheklein: what do we use typer for? explain, plan the mitigation strategy

gto/api.py
shcheklein: This comes from ruff's flake8-bugbear **B905** rule (`zip-without-explicit-strict`), which activates once `requires-python` moves to `>=3.10` (`zip(strict=)` didn't exist on 3.9, so ruff couldn't enforce it before). `ruff --fix` auto-applied the behavior-preserving `strict=False`.

Upgraded to `strict=True` in 50178f4 per the Copilot comment below: both sides of this zip are `2 + len(stages)` items derived from the same `stages` list, so they're provably equal-length and failing fast on future divergence is strictly better.

tests/test_api.py
shcheklein: Same as in `gto/api.py` — ruff B905 (`zip-without-explicit-strict`) kicks in with the 3.10+ target and auto-added the behavior-preserving `strict=False`. Switched to `strict=True` in 50178f4: these are fixed-length parallel lists in an assertion loop, so truncation would mask exactly what the test checks.

pyproject.toml
shcheklein: Verified against current DVC `main` — **no impact**:

- DVC only uses gto's pure library surface: `dvc/repo/artifacts.py` imports `gto.constants` (`assert_name_is_valid`, `fullname_re`), `gto.exceptions`, `gto.tag` (`find`, `parse_tag`), `gto.base` (`sort_versions`). None of those modules import click/typer (that's confined to `gto/cli.py`/`gto/utils.py`), and none changed in this PR.
- DVC still supports and tests py3.9 (`requires-python = ">=3.9"`, 3.9 in its matrix) with `gto>=1.6.0,<2`. Once gto releases with `>=3.10`, pip on py3.9 simply resolves the last 3.9-compatible gto (1.9.x) — installs keep working; 3.9 users just stop receiving gto updates (3.9 has been EOL since Oct 2025).
- No pin conflicts: DVC envs already carry click via celery (which requires `click>=8.1,<9` — same range as the new pin), and nothing in DVC's tree uses typer.
- ⚠️ One requirement: **the next gto release must be `1.10.0`, not `2.0.0`** — DVC caps at `gto<2`, so a major bump would be invisible to DVC until they raise the cap.

Side benefit for DVC docs/users: this PR fixes the `gto` CLI being broken on fresh 3.10+ installs (unpinned typer resolves to 0.26.x → `gto --help` crashes).

pyproject.toml
shcheklein: **What typer is used for:** the entire CLI (`gto/cli.py`) is built on it — `app = Typer(cls=GtoGroup)`, every command declared via typer `Argument`/`Option`, plus `typer.Context`/`typer.Exit`/`typer.colors`. On top of that, gto customizes help/UX by subclassing `typer.core.TyperCommand`/`TyperGroup` **mixed with real-click classes**: `GtoCliMixin(click.Command)` implements grouped help sections ("Read the registry" / "Modify artifacts"), command aliases, and "Examples:" epilogs; `gto_command()` wraps every command with `@click.pass_context` for unified error handling; tests use `typer.testing.CliRunner`.

**Why the pin:** typer 0.24+ replaced its click dependency with a vendored copy (`typer._click`). Two consequences: (1) fresh installs stopped getting click at all while gto imports it directly — the current red `main`; (2) real-click classes mixed into vendored-click `TyperCommand`/`TyperGroup` MROs are incompatible — with typer 0.26 even `gto --help` crashes (`TypeError: Command.__init__() got an unexpected keyword argument 'rich_markup_mode'`). Every fresh `pip install gto` on 3.10+ has shipped a broken CLI since typer 0.24 started resolving.

**Mitigation plan (in preference order):**
1. **Migrate to typer-native APIs** (preferred): `rich_help_panel` covers the command grouping, `epilog` covers the Examples blocks, aliases via registering additional command names; replace the `@click.pass_context` error-handling wrapper with a `ctx: typer.Context` param and typer's re-exported exceptions (`typer.Exit`/`Abort`/`BadParameter`); tests drop `click.testing.Result` for `typer.testing`'s. Then remove both the `typer<0.24` and `click` pins. Scope: essentially one focused PR touching `gto/cli.py` + `tests/conftest.py`.
2. Drop typer for plain click 8.x — the custom machinery is click-level already; typer mainly provides the decorator ergonomics.
3. Rejected: importing from `typer._click` (private API, breaks on any typer minor).

Happy to open a tracking issue for option 1.

gto/api.py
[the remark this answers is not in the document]
shcheklein: Agreed — switched to `strict=True` here in 50178f4. Both iterables are provably parallel (same source/fixed length), so failing fast on divergence is the right semantics. Full suite passes.

tests/test_api.py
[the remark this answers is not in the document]
shcheklein: Agreed — switched to `strict=True` here in 50178f4. Both iterables are provably parallel (same source/fixed length), so failing fast on divergence is the right semantics. Full suite passes.

tests/test_registry.py
```diff
@@ -431,13 +431,15 @@ def test_registry_state_tag_tag(tmp_dir: TmpDir):
         for appeared, expected in zip(
             iter_over(appeared_state["artifacts"][artifact]["versions"]),
             iter_over(expected_state["artifacts"][artifact]["versions"]),
+            strict=False,
```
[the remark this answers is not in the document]
shcheklein: Agreed — switched to `strict=True` here in 50178f4. Both iterables are provably parallel (same source/fixed length), so failing fast on divergence is the right semantics. Full suite passes.

tests/test_registry.py
```diff
@@ -431,13 +431,15 @@ def test_registry_state_tag_tag(tmp_dir: TmpDir):
...
         ):
             check_obj(appeared, expected, exclude["versions"])
 
             for key in ["enrichments", "registrations", "deregistrations"]:
                 for a, e in zip(
                     iter_over(appeared[key]),
                     iter_over(expected[key]),
+                    strict=False,
```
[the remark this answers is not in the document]
shcheklein: Agreed — switched to `strict=True` here in 50178f4. Both iterables are provably parallel (same source/fixed length), so failing fast on divergence is the right semantics. Full suite passes.

pyproject.toml
shcheklein: Correction after checking typer's release notes: the vendoring landed in **typer 0.26.0** (2026-05-26), not 0.24 — 0.24.0 only dropped Python 3.9. Timeline matches our CI: the weekly run started failing June 8, right after 0.26.0's release.

Relaxed the pin accordingly to `typer>=0.4.1,<0.26` in 70bf976 (verified typer 0.25.1 + click 8.4.2: CLI, lints, and the full suite pass).

Typer's stated reasons for vendoring, from the [0.26.0 release notes](https://typer.tiangolo.com/release-notes/): simplifies work for both teams, lets typer evolve independently and enables planned features, and avoids dependency conflicts between packages pinning different click versions. The trade-off is exactly what bit us: click-specific integrations (subclassing `click.Command`, `@click.pass_context`, `click.testing`) are no longer supported against typer's runtime — hence the migration plan above.

pyproject.toml
skshetry: yes, 3.9 is already EOL'd. I don't plan to drop support in dvc though. But 3.9 users should be able to install older version of GTO.

